PR #3, opened 2017-05-09
Asgi transport overhaul

bgreenberg-eb:
Sentinel support has been improved. The transport will now refresh masters from Sentinel on a configurable cooldown, defaulting to 30 seconds. After each refresh, it recreates the channel layer if the masters have changed. It also takes into account the edge cases where Sentinel returns the wrong number of masters or returns masters out of order, both of which would break the channel layer's consistent hash sharding.

This code could be easily integrated into asgi-redis, but the ideal situation would be to patch redis-py to support the notion of "all masters" in its Sentinel class.

### DIFF
--- a/pysoa/common/transport/asgi/core.py
+++ b/pysoa/common/transport/asgi/core.py
@@ -1,10 +1,13 @@
 from __future__ import unicode_literals
 
-import time
 import random
-from redis import StrictRedis
-from asgi_redis import RedisChannelLayer
+import six
+import time
+import logging
+
 import attr
+from asgi_redis import RedisChannelLayer
+import redis
 
 from pysoa.common.transport.exceptions import (
     MessageTooLarge,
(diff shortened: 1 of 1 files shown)


pysoa/common/transport/asgi/core.py
```diff
@@ -19,73 +22,142 @@
...
+            try:
+                redis_client = redis.StrictRedis.from_url(host)
+                master_info_list = redis_client.sentinel_masters()
+                break
+            except redis.ConnectionError as e:
+                connection_errors.append('Failed to connect to {}: {}'.format(host, e))
+                continue
+        if not master_info_list:
+            raise ConnectionError('Could not get master info from sentinel\n{}.'.format('\n'.join(connection_errors)))
+        self._last_sentinel_refresh = time.time()
```
sethbrite: This is just a style thing, but I think this line would be better located if contained within the `_maybe_refresh_masters` `if` statement.

PR #4, opened 2017-05-10
Set ring size on ChannelLayer after creating connection list

bgreenberg-eb:
Fixes a potential bug where the ring size was not correctly set on the channel layer. This was simply an oversight in the initial pass.

I have some ideas of better ways to do this, but this will do what we need and should be good enough for now.


pysoa/common/transport/asgi/core.py
```diff
@@ -96,6 +95,10 @@ def __len__(self):
 
 class SentinelRedisChannelLayer(RedisChannelLayer):
 
+    def __init__(self, *args, **kwargs):
+        super(SentinelRedisChannelLayer, self).__init__(*args, **kwargs)
+        self.ring_size = len(self._connection_list)
+
     def _generate_connections(self, redis_kwargs):
         return SentinelMasterConnectionList(self.hosts, redis_kwargs)
```
sethbrite: Should this method update the ring size as well (i.e. not only on initialization)?

PR #5, opened 2017-05-12
Rewrite local transport to be threadsafe and inlined more

andrewgodwin:
This changes from a pair of classes that share deques to a single class that is both client and server transport, and so can store the information on itself in a threadsafe manner.

It _does_ use a slightly nasty trick to put itself as the server transport, but short of changing how settings work this is hard not to do.

### DIFF
--- a/pysoa/common/transport/local.py
+++ b/pysoa/common/transport/local.py
@@ -11,10 +11,12 @@ from .base import (
 )
 
 
-class ThreadlocalClientTransport(ClientTransport):
+class LocalClientTransport(ClientTransport):
 
     def __init__(self, service_name, server_class_or_path, server_settings):
-        super(ThreadlocalClientTransport, self).__init__(service_name)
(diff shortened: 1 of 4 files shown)


pysoa/common/transport/local.py
```diff
@@ -43,40 +47,61 @@ def __init__(self, service_name, server_class_or_path, server_settings):
             'path': 'pysoa.common.transport.local:ThreadlocalServerTransport',
         }
 
+        # Set an empty queued request; we'll use this later
+        self._current_request = None
+
+        # Set up a deque for responses for just this client
+        self.response_messages = deque([])
```
vartec: maybe set `maxlen` to better emulate what ASGI is doing?

andrewgodwin: In ASGI the client would start consuming the responses before the server had finished them, though. In reality we'd never have a max len any code would hit.

PR #7, opened 2017-07-26
Refactor Client to include routing functionality, and remove Router

bgreenberg-eb:
Merge the functionality of the Router and Client, so that the user no longer needs to initialize a Router and call `get_client` before calling actions.

The Client now stores mini-clients, called Handlers, that contain the transport, serializer and middleware for each service. All calling methods now take a service name as the first positional argument and use the handler corresponding to that service name. Handlers are lazy-loaded when the service is first called. I've also refactored the Client code slightly for clarity.

You can now have one client that can make both remote and local service calls. This can greatly simplify client creation and make settings smaller. It may still be better to have separate clients for remote and local services in some circumstances, but now we have the choice.

Tests have been updated to reflect the new Client interface. Since expansions are now on `call_actions`, I added another test for expansions on multiple actions. I also updated all of the test files that I touched to use `TestCase`-based tests, in accordance with our agreements on test style.


tests/common_tests/test_test.py
```diff
@@ -1,31 +1,32 @@
 from __future__ import unicode_literals
 
-from pysoa.test.stub_service import StubClient
+# from pysoa.client.client import ServiceHandler
```
vartec: remove

pysoa/client/settings.py
```diff
@@ -1,38 +1,8 @@
...
-    defaults = {
-        'client': {'path': 'pysoa.client:Client'},
-        'cacheable': False,
-    }
-
-    def convert_client(self, value):
-        return self.standard_convert_path(value)
-
-
-class ASGIClientSettings(ClientSettings):
```
vartec: So why isn't this schema needed anymore? Has ASGI configuration been moved somewhere else that's not in this diff?

bgreenberg-eb: The ASGI transport itself enforces a schema for its arguments via `attrs`, so adding an extra level of protection via a settings schema did not seem to be worth the extra complexity.

bgreenberg-eb:
I'm in the process of updating this to implement some ideas I had over the weekend, so this should be considered WIP right now.

PR #8, opened 2017-08-02
Refactor Client to include routing functionality, and remove Router

bgreenberg-eb:
## Overview

Merge the functionality of the `Router` and `Client`, so that the user no longer needs to initialize a `Router` and call `get_client` before calling actions. The `Client` now stores mini-clients, called `ServiceHandler`s, that incorporate the core request and response functionality for single services.`ServiceHandler`s are lazy-loaded when the service is first called.

## API changes

All calling methods now take a service name as the first positional argument and use the `ServiceHandler` corresponding to that service name. 

Because `ServiceHandler`s are self-contained and independent, one `Client` can now make both remote and local service calls. This can greatly simplify client creation and make settings smaller.

`Client.send_request` signature now matches that of `call_actions` (minus expansions), to make the synchronous and asynchronous APIs more consistent. 

## Test plan

Tests have been updated to reflect the new `Client` interface. Since expansions are now on `call_actions`, I added another test for expansions on multiple actions. I also updated all of the test files that I touched to use `TestCase`-based tests, in accordance with our agreements on test style.

## Other notes
Removed the `.arclint` file and added a `.pep8` file.


pysoa/client/client.py
```diff
@@ -4,22 +4,142 @@
...
-    def __init__(self, service_name, transport, serializer, middleware=None, context=None):
-        self.service_name = service_name
-        self.transport = transport
-        self.serializer = serializer
-        self.middleware = middleware or []
-        self.context = context
-        self.request_counter = 0
+    settings_class = ASGIClientSettings
+    handler_class = ServiceHandler
+
```
sethbrite: We should probably have `config` schema validation.

pysoa/client/client.py
```diff
@@ -54,11 +174,153 @@ def __str__(self):
...
+            for type_node in self.expansion_converter.dict_to_trees(expansions):
+                for action in response.actions:
+                    exp_objects = type_node.find_objects(action.body)
+                    objs_to_expand.extend(
+                        (exp_object, type_node.expansions)
+                        for exp_object in exp_objects
+                    )
+
+            # Loop until we have no outstanding requests or responses
+            while objs_to_expand or any(requests for requests in exp_service_requests.values()):
```
sethbrite: Just curious, what happened to the `six` here and in other iterations in this file?

pysoa/client/client.py
bgreenberg-eb: The settings class does that. Do you mean *more* schema validation, as in the old `ASGITransportSchema`?

pysoa/client/client.py
bgreenberg-eb: It seemed that they weren't necessary. I tested this code path in both 2.7 and 3.5 and it worked as intended.

pysoa/client/client.py
```diff
@@ -54,11 +174,153 @@ def __str__(self):
...
+        if response_id != request_id:
+            raise Exception('Got response with ID {} for request with ID {}'.format(response_id, request_id))
+        if response.errors:
+            raise self.JobError(response.errors)
+        error_actions = [action for action in response.actions if action.errors]
+        if error_actions:
+            raise self.CallActionError(error_actions)
+
+        # Perform expansions
+        if expansions and hasattr(self, 'expansion_converter'):
```
andrewgodwin: Should we handle the error case (requesting expansions and there not being a converter to handle them)? Right now it silently fails.

bgreenberg-eb: Probably. I consider that part of the larger issue of how we handle errors in expansions, though, and I'd like to handle that in a separate PR.

andrewgodwin: Alright. As long as we cover it somewhere I'm happy.

pysoa/client/client.py
sethbrite: A `config` is basically the routing information for each service, correct? Right now, it appears that each client is being initialized with `ASGIClientSettings` by default. Does this work for local services? The client now acts as the router, so you should be able to call any service for which transport and serialization information has been configured.

My original request for `config` schema validation was meant to catch errors in the `config` dictionary format (i.e. not `{<service name key>: <client config dict>, ...}`).

bgreenberg-eb: The idea there was to make the Client default to using ASGI. `ASGIClientSettings` just contains a default for the Transport class, and is otherwise just the standard SOA settings schema.

I removed the schema validation for ASGI transports because it was redundant; the Transport core itself handles settings validation. However, the advantage of the schema validation was that it would fail at Client initialization rather than Handler initialization. I don't feel strongly about this either way, so I can go back to using the old `ASGIClientSettings` if this seems like a bad idea.

bgreenberg-eb: I could see having a dynamic settings class, or even just one with multiple schema `OR`'d together into one super-schema. I think that deserves its own PR though.

sethbrite: Alright, this seems fine.

PR #17, opened 2017-10-17
Add exc_info option to logger.error statements

bgreenberg-eb:
This will still cause the logger to include traceback information in log lines, while allowing different loggers to handle tracebacks their way.

The use case that prompted this was the `SentryHandler` from `raven`, which cannot correctly format exception information in Sentry requests without `exc_info`.

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -171,7 +171,7 @@ class Server(object):
         except Exception:
             error_str, traceback_str = "Error formatting error", traceback.format_exc()
         # Log what happened
-        self.logger.exception(error, exc_info=True)
+        self.logger.exception(error)
         # Make a barebones job response
         error_dict = {
             'code': ERROR_CODE_SERVER_ERROR,
@@ -288,7 +288,7 @@ class Server(object):
                 # Get, process, and execute the next JobRequest
                 self.handle_next_request()
         except Exception:
-            self.logger.exception("Unhandled server error", exc_info=True)
+            self.logger.exception("Unhandled server error")
         finally:
             self.logger.info("Server shutting down")
 
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -171,7 +171,7 @@ class Server(object):
         except Exception:
             error_str, traceback_str = "Error formatting error", traceback.format_exc()
         # Log what happened
-        self.logger.error("Unhandled error: %s", traceback_str)
+        self.logger.error(error, exc_info=True)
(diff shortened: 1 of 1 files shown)


pysoa/server/server.py
```diff
@@ -288,7 +288,7 @@ def run(self):
                 # Get, process, and execute the next JobRequest
                 self.handle_next_request()
         except Exception:
-            self.logger.error("Unhandled server error: %s", traceback.format_exc())
+            self.logger.error("Unhandled server error", exc_info=True)
```
bar: the correct way to log exceptions is to call `.exception` like `self.logger.exception("Unhandled server error")`

which _MUST_ be called within the context of an exception

pysoa/server/server.py
```diff
@@ -171,7 +171,7 @@ def handle_error(self, error, variables=None):
         except Exception:
             error_str, traceback_str = "Error formatting error", traceback.format_exc()
         # Log what happened
-        self.logger.error("Unhandled error: %s", traceback_str)
+        self.logger.error(error, exc_info=True)
```
nickwilliams-eventbrite: This isn't going to work as expected, because it's no longer within the scope of the caught exception. `sys.exc_info()` will return `None, None, None` once the exception is no longer being handled. It's fine below, but not here. Here, you'll need to formulate an exception tuple (`exception_type, exception_value, traceback`) and pass that to `exc_info` like so:

    try:
        exc_tuple = [exception_type], [exception_value], traceback.format_exc()
        ...
    except Exception as e:
        exc_tuple = type(e), e, traceback.format_exc()
        ...
    self.logger.error(error, exc_info=exc_tuple)

pysoa/server/server.py
nickwilliams-eventbrite: I agree with Ber, here. Use `self.logger.exception`.

pysoa/server/server.py
bgreenberg-eb: I wasn't sure whether or not it would work when I first tried it, but it does. Since `handle_error` is only ever called within an exception context, `sys.exc_info()` returns the traceback as expected.

I also did a sanity check by raising an exception inside an action and adding `print sys.exc_info()` at line 173, and got an exception tuple as expected. It really does work that way.

nickwilliams-eventbrite: TIL

bgreenberg-eb:
Removed `exc_info=True` from `exception()` calls as requested by @bar.

PR #20, opened 2017-11-03
Add support for recording metrics directly within SOA clients, servers, and transports

nickwilliams-eventbrite:
- Creates an interface for generic metrics recording
- Specifies a new configuration schema for setting up metrics that will be compatible with our current internal uses
- Integrates the recording of metrics into the server, client, and Redis transport. No effort is being expended on metricizing the ASGI transport, since it is being deprecated. No effort is being expended on metricizing the Local transport, as it has essentially no overhead compared to the client, which is metricized.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -536,10 +536,59 @@ Transport exceptions
 - ``MessageReceiveError``: The transport encountered any other error while trying to receive a message.
 
 
-ASGI Transport
-++++++++++++++
+Redis Gateway Transport
++++++++++++++++++++++++
 
-The ``transport.asgi`` module provides a transport implementation that uses the `ASGI <http://channels.readthedocs.io/en/stable/asgi.html>`_ protocol. This is the recommended transport for use with ``pysoa``, as it provides a convenient and performant backend for asynchronous service requests. It is also the technology underlying `Django channels <https://channels.readthedocs.io/en/stable/>`_, so you know it's good.
+The ``transport.redis_gateway`` module provides a transport implementation that uses Redis (in simple or Sentinel mode)
+for sending and receiving messages. This is the recommended transport for use with ``pysoa``, as it provides a
+convenient and performant backend for asynchronous service requests.
+
+Standard and Sentinel modes
+***************************
+
+The Redis Gateway transport has two primary modes of operation: in "standard" mode, the channel layer will connect to a
+specified list of Redis hosts, while in "Sentinel" mode, the channel layer will connect to a list of Sentinel hosts and
+use Sentinel to find its Redis hosts.
+
+Configuration
+*************
+
+The Redis Gateway transport takes the following extra keyword arguments for configuration:
+
+- ``backend_type``: Either "redis.standard" or "redis.sentinel" to specify which Redis backend to use (required)
+
+- ``backend_layer_kwargs``: A dictionary of arguments to pass to the backend layer
+
+  + ``connection_kwargs``: A dictionary of arguments to pass to the underlying Redis client
+
+  + ``hosts``: A list of strings (host names / IP addresses) or tuples (host names / IP addresses and ports) for Redis hosts or sentinels to which to connect (will use "localhost" by default)
+
+  + ``redis_db``: The Redis database number to use (a shortcut for specifying ``connection_kwargs['db']``)
+
+  + ``redis_port``: The connection port to use (a shortcut for providing this for every entry in ``hosts``
+
+  + ``sentinel_refresh_interval``: How often, in seconds, master/slave data should be refreshed from the Sentinel (only for type "redis.sentinel") (refreshes every connection by default)
+
+  + ``sentinel_services``: Which Sentinel services to use (only for type "redis.sentinel") (will be auto-discovered from the Sentinel by default)
+
+- ``message_expiry_in_seconds``: How long a message may remain in the queue before it is considered expired and discarded
+
+- ``queue_capacity``: The maximum number of messages a given Redis queue may hold before the transport should stop pushing messages to it (defaults to 10,000)
+
+- ``queue_full_retries``: The number of times the transport should retry sending to a Redis queue that is at capacity before it raises an error and stops trying (defaults to 10)
+
+- ``receive_timeout_in_seconds``: How long the transport should block waiting to receive a message before giving up (on the server, this controls how often the server request-process loops; on the client, this controls how long before it raises an error for waiting too long for a response)
+
+- ``serializer_config``: A standard serializer configuration as described in `Serializer configuration`_ (defaults to Msgpack)
+
+
+ASGI Transport (DEPRECATED)
++++++++++++++++++++++++++++
+
+**WARNING:** The ASGI transport is deprecated due to performance issues. It will be removed before the first 1.0.0 GA
+release. Please migrate your projects to the `Redis Gateway Transport`_.
+
+The ``transport.asgi`` module provides a transport implementation that uses the `ASGI <http://channels.readthedocs.io/en/stable/asgi.html>`_ protocol. It is also the technology underlying `Django channels <https://channels.readthedocs.io/en/stable/>`_.
 
 The reference ASGI implementation, used in ``transport.asgi``, uses Redis as a message backend. The protocol is backend-agnostic, however; if you need to use a different backend, you can implement your own ``ASGITransportCore`` or extend the existing one.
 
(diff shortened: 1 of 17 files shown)


pysoa/common/schemas.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import unicode_literals
```
nickwilliams-eventbrite: The contents of this file were simply moved from pysoa/common/settings.py to resolve a circular import.

pysoa/server/server.py
```diff
@@ -43,7 +43,7 @@ class Server(object):
             to Action subclasses.
     """
 
-    settings_class = ServerSettings
+    settings_class = PolymorphicServerSettings
```
nickwilliams-eventbrite: Something missed from previous PR.

pysoa/test/stub_service.py
```diff
@@ -7,16 +7,19 @@
...
 from pysoa.server.action import (
     Action,
     ActionError,
 )
 
 
-def _make_stub_action(action_name, body={}, errors=[]):
+def _make_stub_action(action_name, body=None, errors=None):
+    body = body or {}
+    errors = errors or {}
```
nickwilliams-eventbrite: Wowzers! How did this bug never bite us before?? 😬

nickwilliams-eventbrite:
@ebpitts, I am especially interested in your review on this PR. I'd love to know if you feel these metrics will give us sufficient insight into PySOA performance.

pysoa/common/transport/base.py
```diff
@@ -9,15 +9,16 @@
...
-types, the metadata may include a mimetype to tell the endpoint receiving the message
+types, the metadata may include a mime type to tell the endpoint receiving the message
 which type of serializer to use.
 """
 
 
 class ClientTransport(object):
 
-    def __init__(self, service_name):
+    def __init__(self, service_name, metrics):
```
bgreenberg-eb: I would make `metrics` a keyword arg that default to `NoOpMetricsRecorder`, just for consistency. In `ServerTransport` as well.

pysoa/test/stub_service.py
bgreenberg-eb: It never caused any problems because nothing mutates those values. But this is more correct for sure.

ebpitts:
@nickwilliams-eventbrite to faciliate review, could you document each metric and its meaning? Either here in the review or, even better, in your docs/ directory.

nickwilliams-eventbrite:
@ebpitts, I have added documentation as requested: https://github.com/eventbrite/pysoa/blob/add-metrics/docs/api.rst#metrics

docs/api.rst
```diff
@@ -536,10 +536,59 @@ Transport exceptions
...
+
+  + ``redis_port``: The connection port to use (a shortcut for providing this for every entry in ``hosts``
+
+  + ``sentinel_refresh_interval``: How often, in seconds, master/slave data should be refreshed from the Sentinel (only for type "redis.sentinel") (refreshes every connection by default)
+
+  + ``sentinel_services``: Which Sentinel services to use (only for type "redis.sentinel") (will be auto-discovered from the Sentinel by default)
+
+- ``message_expiry_in_seconds``: How long a message may remain in the queue before it is considered expired and discarded
+
+- ``queue_capacity``: The maximum number of messages a given Redis queue may hold before the transport should stop pushing messages to it (defaults to 10,000)
```
ebpitts: That's a heck of a default! I would have gone with something closer to 100. Thinking of the clients making requests to the servers, I's think we just need some small multiple of the number of workers we expect to run in order to smooth out bursts.

This also applies to the server responding to a client though, doesn't it? The value we set here determines how many unread responses a client can have from a single server?

docs/api.rst
```diff
@@ -536,10 +536,59 @@ Transport exceptions
...
+
+  + ``connection_kwargs``: A dictionary of arguments to pass to the underlying Redis client
+
+  + ``hosts``: A list of strings (host names / IP addresses) or tuples (host names / IP addresses and ports) for Redis hosts or sentinels to which to connect (will use "localhost" by default)
+
+  + ``redis_db``: The Redis database number to use (a shortcut for specifying ``connection_kwargs['db']``)
+
+  + ``redis_port``: The connection port to use (a shortcut for providing this for every entry in ``hosts``
+
+  + ``sentinel_refresh_interval``: How often, in seconds, master/slave data should be refreshed from the Sentinel (only for type "redis.sentinel") (refreshes every connection by default)
```
ebpitts: You use a persistent connection instead of reopening a connection for every operation, right? Does this mean the default is to refresh every time a connection is opened or every time a connection is used?

Do you refresh the connections when a redis or network related exception occurs? That's what is most important, not periodically refreshing, imho.

docs/api.rst
```diff
@@ -536,10 +536,59 @@ Transport exceptions
...
+
+  + ``sentinel_services``: Which Sentinel services to use (only for type "redis.sentinel") (will be auto-discovered from the Sentinel by default)
+
+- ``message_expiry_in_seconds``: How long a message may remain in the queue before it is considered expired and discarded
```
ebpitts: Does this have a default?

docs/api.rst
```diff
@@ -536,10 +536,59 @@ Transport exceptions
...
+
+- ``queue_capacity``: The maximum number of messages a given Redis queue may hold before the transport should stop pushing messages to it (defaults to 10,000)
 
-The ``transport.asgi`` module provides a transport implementation that uses the `ASGI <http://channels.readthedocs.io/en/stable/asgi.html>`_ protocol. This is the recommended transport for use with ``pysoa``, as it provides a convenient and performant backend for asynchronous service requests. It is also the technology underlying `Django channels <https://channels.readthedocs.io/en/stable/>`_, so you know it's good.
+- ``queue_full_retries``: The number of times the transport should retry sending to a Redis queue that is at capacity before it raises an error and stops trying (defaults to 10)
+
+- ``receive_timeout_in_seconds``: How long the transport should block waiting to receive a message before giving up (on the server, this controls how often the server request-process loops; on the client, this controls how long before it raises an error for waiting too long for a response)
```
ebpitts: Does this have a default?

jshuping:
Thanks a ton for doing this. Seems like a very thorough set of metrics!

docs/api.rst
```diff
@@ -536,10 +536,59 @@ Transport exceptions
...
+
+  + ``sentinel_services``: Which Sentinel services to use (only for type "redis.sentinel") (will be auto-discovered from the Sentinel by default)
+
+- ``message_expiry_in_seconds``: How long a message may remain in the queue before it is considered expired and discarded
+
+- ``queue_capacity``: The maximum number of messages a given Redis queue may hold before the transport should stop pushing messages to it (defaults to 10,000)
 
-The ``transport.asgi`` module provides a transport implementation that uses the `ASGI <http://channels.readthedocs.io/en/stable/asgi.html>`_ protocol. This is the recommended transport for use with ``pysoa``, as it provides a convenient and performant backend for asynchronous service requests. It is also the technology underlying `Django channels <https://channels.readthedocs.io/en/stable/>`_, so you know it's good.
+- ``queue_full_retries``: The number of times the transport should retry sending to a Redis queue that is at capacity before it raises an error and stops trying (defaults to 10)
```
ebpitts: A few thoughts:

1) Should we retry by default at all in the transport layer? Or should we have a retry policy in a higher level (middleware?) that can make retry decisions both on transport errors and on error responses.

2) If we do retry here, should the retry only occur on a full queue, or should it retry on other transport errors like redis connection timing out?

3) What is the delay between retry attempts? Is it constant?

4) 10 seems really high for number of retries by default. I'd lean towards not retrying by default at all.

docs/api.rst
```diff
@@ -536,10 +536,59 @@ Transport exceptions
...
+
+The Redis Gateway transport takes the following extra keyword arguments for configuration:
+
+- ``backend_type``: Either "redis.standard" or "redis.sentinel" to specify which Redis backend to use (required)
+
+- ``backend_layer_kwargs``: A dictionary of arguments to pass to the backend layer
+
+  + ``connection_kwargs``: A dictionary of arguments to pass to the underlying Redis client
+
+  + ``hosts``: A list of strings (host names / IP addresses) or tuples (host names / IP addresses and ports) for Redis hosts or sentinels to which to connect (will use "localhost" by default)
```
ebpitts: What is the timeout used when we attempt to connect?

nickwilliams-eventbrite: This can be specified in `connection_kwargs` and is a detail left up to the documentation for the Redis-Py library (I have updated this to reference the Redis-Py library).

docs/api.rst
nickwilliams-eventbrite: Yes, a persistent connection is used, from a pool. We get a connection from the pool on every Redis call. For sending to a server or receiving from a client, this is just round-robin whichever connection we get next. For sending to a client or receiving from a server, this is consistently hashed (always the same connection for a given client ID). The default is to refresh every time we get a connection from the pool, not every time we open a new connection.

We do not refresh the connections at this layer when a Redis- or network-related exception occurs. On the client side, that request fails and the client is re-created for the next request, which results in the same effect as refreshing connections. On the server side, I'm not sure how we handle it. But I would note that this does not reflect a change in behavior from the current behavior (and, in this PR, I'm just documenting this; the new transport was added in an earlier PR).

docs/api.rst
nickwilliams-eventbrite: Yes: Sixty seconds, and I have updated the documentation.

docs/api.rst
nickwilliams-eventbrite: The default applies to both the queues clients use to send requests to servers and the queues servers use to respond to clients. This is comparable to the existing behavior, and serves as an argument for defining non-default settings overrides.

docs/api.rst
nickwilliams-eventbrite: This retry ONLY applies to queue-full errors. I firmly believe we should retry this, and we should do so for some time before giving up. Taking a little longer is preferable to erring. We do not retry on any other errors. If we wanted to do so, we would handle that as a separate feature request and approach it when made.

docs/api.rst
nickwilliams-eventbrite: It does: 5 seconds, and I have updated the documentation to say this. We may decide to make this a little higher on the client (even as much as 10) and a little lower on the server (even as little as 1).

PR #26, opened 2017-12-02
Add option for PySOA server to gracefully recover from Redis master failover

nickwilliams-eventbrite:


pysoa/common/transport/redis_gateway/backend/sentinel.py
```diff
@@ -117,7 +135,18 @@ def _get_connection(self, index=None):
                 )
             )
 
-        return self._master_for(self._services[index])
+        for i in range(-1, self._sentinel_failover_retries):
```
andrewgodwin: It's nicer to write this as `for i in range(self._sentinel_failover_retries + 1)` IMO.

pysoa/common/transport/redis_gateway/backend/sentinel.py
```diff
@@ -117,7 +135,18 @@ def _get_connection(self, index=None):
...
+            try:
+                return self._master_for(self._services[index])
+            except redis.sentinel.MasterNotFoundError:
+                self._last_sentinel_refresh = 0  # make sure we reach out to get master info again on next call
+                if i == self._sentinel_failover_retries - 1:
```
andrewgodwin: Use `for ... else` with a `break` on success rather than this, please!

pysoa/common/transport/redis_gateway/backend/sentinel.py
```diff
@@ -117,7 +135,18 @@ def _get_connection(self, index=None):
...
             )
 
-        return self._master_for(self._services[index])
+        for i in range(self._sentinel_failover_retries + 1):
+            try:
+                return self._master_for(self._services[index])
+            except redis.sentinel.MasterNotFoundError:
+                self._last_sentinel_refresh = 0  # make sure we reach out to get master info again on next call
+                self._get_counter('backend.sentinel.master_not_found_retry').increment()
+                time.sleep((1 ** i + random.random()) / 4.0)
```
ebpitts: What range are you trying to produce on each iteration?

Won't `1 ** i` always equal 1?

ebpitts:
Why limit retries? What happens to the process after we've exceeded the limit?

pysoa/common/transport/redis_gateway/backend/sentinel.py
nickwilliams-eventbrite: Ooops. That slipped through. Should be 2:

```
In [6]: for i in range(-1, 10):
    print((2 ** i + random.random()) / 4.0)
   ...:     
0.185900067591
0.259383624881
0.666199774509
1.17092891973
2.0800915192
4.21767195389
8.14190244716
16.2083223142
32.1969619394
64.0182664696
128.177728188
```

I'll get a new PR posted to fix this.

nickwilliams-eventbrite:
>Why limit retries? What happens to the process after we've exceeded the limit?
There are any number of reasons we might want to limit retries. If we retry indefinitely, it's not necessarily brought to anyone's attention. If we exceed the limit, it propagates up to the server, which logs the error and quits. Then nomad restarts it. This gets people's attention. _**Importantly, this is a setting that Ops will have control over, so y'all can fine-tune it to your discretion.**_

ebpitts:
Does the retry behavior implemented here apply only to the server, or also to the client? The PR title says server but looks like changes to client code as well to me. Thinking of Eventbrite's use case, we definitely don't want core to enter a crash loop because it can't talk to geo service. We also don't want core (or other clients) to `sleep()` waiting to refresh masters and reinitialize redis connection. That will block core from handling other requests that it might be able to process even without this connection  I agree it's desirable to throttle this, so that we don't overwhelm redis-sentinel, but client-side that should be implemented as a circuit-breaker that immediately fails requests during the duration that we are waiting, instead of literally waiting via `sleep()`.

nickwilliams-eventbrite:
>Does the retry behavior implemented here apply only to the server, or also to the client? The PR title says server but looks like changes to client code as well to me. Thinking of Eventbrite's use case, we definitely don't want core to enter a crash loop because it can't talk to geo service. We also don't want core (or other clients) to sleep() waiting to refresh masters and reinitialize redis connection. That will block core from handling other requests that it might be able to process even without this connection I agree it's desirable to throttle this, so that we don't overwhelm redis-sentinel, but client-side that should be implemented as a circuit-breaker that immediately fails requests during the duration that we are waiting, instead of literally waiting via sleep().

Ops has total control over this. The default setting, which is 0 retries, means that it will fail fast ("circuit breaker"), just as you describe. In a client, this means it will just fail right away (only a server would enter the restart loop). However, previously it was not `sleep`-ing in this case. The change that @andrewgodwin asked for (moving `raise` out of the loop) resulted in its `sleep`-ing even on fast-fail, and also resulted in incorrect metrics. PR to fix this: https://github.com/eventbrite/pysoa/pull/29

PR #27, opened 2017-12-02
Add support for a cached client transport to increase connection re-use

nickwilliams-eventbrite:


ebpitts:
Caching the client transport raises the question of how it handles connection errors, such as when the during a failover. IIRC we were caching the client in the past, but it did not handle that, so as a quick hack we switched to always establishing a new connection.

nickwilliams-eventbrite:
>Caching the client transport raises the question of how it handles connection errors, such as during a planned or automated failover. IIRC we were caching the client in the past, but it did not handle that, so as a quick hack we switched to always establishing a new connection.

This is the next step following that "quick hack." Under the default PySOA Sentinel settings, if transport caching is enabled, the client will still call the `SENTINEL MASTERS` command on every request, so all requests will fail for a few seconds when the failover is in progress (just like they would now), and then the cached transports will pick up the new master information. How long transports are cached is a setting Ops has control over (defaults off), so you could make it a mere 10 or 15 seconds if you're unsure about this `SENTINEL MASTERS` failover working, or a longer 30 or 60 seconds if you're as confident in it as am I.

Eventually, I would like to enable `sentinel_refresh_interval` for the client settings to improve performance there (it's disabled right now; enabling it would reduce calls to `SENTINEL MASTERS`). Due to the previous pull request which landed, even with `sentinel_refresh_interval` enabled, a master error would cause the client transport to reset and call `SENTINEL MASTERS` for the next request.

PR #33, opened 2017-12-19
Use master_for correctly to reduce number of Redis connections

nickwilliams-eventbrite:
For other Redis-consuming projects, we have traditionally used the `sentinel_sharded_redis` library, which has a `sentinel_refresh_interval` setting. This setting controls how often `sentinel_sharded_redis` reaches out to Sentinel to get the latest information about Redis masters.

When consuming the `redis-py` library in this project, we applied the same pattern, adding a `sentinel_refresh_interval` setting to the transport and using that to control how often `redis-py`'s `sentinel.master_for(...)` is called. This is not the way `sentinel.master_for(...)` was designed to be used.

`sentinel.master_for(...)` creates a whole new `StrictRedis` client with a whole new connection pool every time it is called. We are wasting connections that remain open and unused until garbage collection runs, and opening connections too often, by regularly calling `sentinel.master_for(...)`. This method was designed to be called in a startup or lazy-startup capacity, and the client it returns should be used indefinitely. The client it returns uses a special managed connection pool with special managed connections that keeps track of master information and automatically closes and re-opens connections if the master information changes.

In this change, we are removing the `sentinel_refresh_interval` setting (but it can still be included for a no-op for temporary backwards compatibility with existing settings). We will no longer call `sentinel.master_for(...)` regularly. Instead, we call it the first time we need a master client, and then we use that master client indefinitely.

- `master_for` behavior: https://github.com/andymccurdy/redis-py/blob/5109cb4f6b610e8d5949716a16435afbbf35075a/redis/sentinel.py#L246-L272
- Sentinel-managed connection pool behavior: https://github.com/andymccurdy/redis-py/blob/5109cb4f6b610e8d5949716a16435afbbf35075a/redis/sentinel.py#L99-L108
- Sentinel-managed connection behavior: https://github.com/andymccurdy/redis-py/blob/5109cb4f6b610e8d5949716a16435afbbf35075a/redis/sentinel.py#L33-L66
- An example of Flask using `redis-py` Sentinel the same way we're proposing in this change: https://github.com/exponea/flask-redis-sentinel/blob/9f0003345d3756c19ce913acb0e25b613f8368a6/flask_redis_sentinel.py#L62-L77

### DIFF
--- a/pysoa/common/transport/redis_gateway/core.py
+++ b/pysoa/common/transport/redis_gateway/core.py
@@ -28,6 +28,7 @@ from pysoa.common.transport.redis_gateway.constants import (
     REDIS_BACKEND_TYPE_SENTINEL,
     REDIS_BACKEND_TYPES,
 )
+from pysoa.utils import dict_to_hashable
 
 
 logger = logging.getLogger('pysoa.common.transport')
@@ -41,6 +42,12 @@ def valid_backend_type(_, __, value):
 @attr.s()
 class RedisTransportCore(object):
     """Handles communication with Redis."""
+
+    # The backend layer holds Redis connections and should be reused as much as possible to reduce Redis connections.
+    # Given identical input settings, two given backend layer instances will operate identically, and so we cash using
+    # input variables as a key. This applies even across services--backend layers have no service-specific code, so
+    # a single backend can be used for multiple services if those services' backend settings are the same.
+    _backend_layer_cache = {}
 
     backend_type = attr.ib(validator=valid_backend_type)
 
@@ -121,14 +128,19 @@ class RedisTransportCore(object):
     @property
     def backend_layer(self):
         if self._backend_layer is None:
-            backend_layer_kwargs = deepcopy(self.backend_layer_kwargs)
-            with self._get_timer('backend.initialize'):
-                if self.backend_type == REDIS_BACKEND_TYPE_SENTINEL:
-                    self._backend_layer = SentinelRedisClient(**backend_layer_kwargs)
-                    self._backend_layer.metrics_counter_getter = lambda name: self._get_counter(name)
-                else:
-                    self._backend_layer = StandardRedisClient(**backend_layer_kwargs)
+            cache_key = (self.backend_type, dict_to_hashable(self.backend_layer_kwargs))
+            if cache_key not in self._backend_layer_cache:
+                with self._get_timer('backend.initialize'):
+                    backend_layer_kwargs = deepcopy(self.backend_layer_kwargs)
+                    if self.backend_type == REDIS_BACKEND_TYPE_SENTINEL:
+                        self._backend_layer_cache[cache_key] = SentinelRedisClient(**backend_layer_kwargs)
+                    else:
+                        self._backend_layer_cache[cache_key] = StandardRedisClient(**backend_layer_kwargs)
 
+            self._backend_layer = self._backend_layer_cache[cache_key]
+
+        # Each time the backend layer is accessed, use _this_ transport's metrics recorder for the backend layer
+        self._backend_layer.metrics_counter_getter = lambda name: self._get_counter(name)
         return self._backend_layer
 
     # noinspection PyAttributeOutsideInit
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -576,8 +576,6 @@ The Redis Gateway transport takes the following extra keyword arguments for conf
 
   + ``sentinel_failover_retries``: How many times to retry (with a delay) getting a connection from the Sentinel when a master cannot be found (cluster is in the middle of a failover) (only for type "redis.sentinel") (fails on the first error by default)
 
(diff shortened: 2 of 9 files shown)


pysoa/common/transport/redis_gateway/core.py
```diff
@@ -121,14 +127,19 @@ def __attrs_post_init__(self):
...
         if self._backend_layer is None:
-            backend_layer_kwargs = deepcopy(self.backend_layer_kwargs)
-            with self._get_timer('backend.initialize'):
-                if self.backend_type == REDIS_BACKEND_TYPE_SENTINEL:
-                    self._backend_layer = SentinelRedisClient(**backend_layer_kwargs)
-                    self._backend_layer.metrics_counter_getter = lambda name: self._get_counter(name)
-                else:
-                    self._backend_layer = StandardRedisClient(**backend_layer_kwargs)
-
+            cache_key = repr((self.backend_type, self.backend_layer_kwargs))
```
vartec: nitpick: `self.backend_layer_kwargs` is a dict, it will probably work fine because even in Py2 within thread repr is consistent, but do we want to rely on undocumented behavior?

nickwilliams-eventbrite: For two or more dicts with the same content, `repr(each_dict)` is guaranteed to be the same within a given Python process. It's only guaranteed to be the same across processes in Python 3, but in Python 2 it's guaranteed to be the same within the same process, which is all we need.

nickwilliams-eventbrite: I turned out to be incorrect here. Fix incoming.

nickwilliams-eventbrite: For reference:

```
>>> d1 = {u'hosts': [u'1.2.3.4', u'5.6.7.8', u'9.10.11.12'], u'connection_kwargs': {u'password': u'abc123meta', u'socket_connect_timeout': 20}, u'sentinel_services': [u'service1', u'service2'], u'sentinel_failover_retries': 6}
>>> d2 = {u'sentinel_services': [u'service1', u'service2'], u'sentinel_failover_retries': 6, u'connection_kwargs': {u'password': u'abc123meta', u'socket_connect_timeout': 20}, u'hosts': [u'1.2.3.4', u'5.6.7.8', u'9.10.11.12']}
>>> d3 = {u'connection_kwargs': {u'socket_connect_timeout': 20, u'password': u'abc123meta'}, u'sentinel_services': [u'service1', u'service2'], u'sentinel_failover_retries': 6, u'hosts': [u'1.2.3.4', u'5.6.7.8', u'9.10.11.12']}
>>> d1 == d2
True
>>> d1 == d3
True
>>> repr(d1) == repr(d2)
False
>>> repr(d1) == repr(d3)
True
```

But:

```
>>> %timeit repr(d1)
The slowest run took 5.76 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 3.48 µs per loop

>>> %timeit dict_to_hashable(d1)
The slowest run took 4.16 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 4.07 µs per loop
```

And the result of `dict_to_hashable` can be used directly as a dict key.

ebpitts:
> The client it returns uses a special managed connection pool with special managed connections that keeps track of master information and automatically closes and re-opens connections if the master information changes.

Can you link to the relevant code in redis-py?

We'll want to retest the client and server behavior during failover after this change lands.

nickwilliams-eventbrite:
@ebpitts:

- `master_for` behavior: https://github.com/andymccurdy/redis-py/blob/5109cb4f6b610e8d5949716a16435afbbf35075a/redis/sentinel.py#L246-L272
- Sentinel-managed connection pool behavior: https://github.com/andymccurdy/redis-py/blob/5109cb4f6b610e8d5949716a16435afbbf35075a/redis/sentinel.py#L99-L108
- Sentinel-managed connection behavior: https://github.com/andymccurdy/redis-py/blob/5109cb4f6b610e8d5949716a16435afbbf35075a/redis/sentinel.py#L33-L66
- An example of Flask using `redis-py` Sentinel the same way we're proposing in this change: https://github.com/exponea/flask-redis-sentinel/blob/9f0003345d3756c19ce913acb0e25b613f8368a6/flask_redis_sentinel.py#L62-L77

pseudoben:
Outstanding.

nickwilliams-eventbrite:
Results of load testing in Stage:
- Whereas "new Redis connections/sec" was ~1,800 without `sentinel_refresh_interval`/`transport_cache_time_in_seconds` and ~10 with it, it's ~1 with this new pull request.
- Whereas "persistent Redis connections" was ~50 without `sentinel_refresh_interval`/`transport_cache_time_in_seconds` and ~1,500 and jagged with it, it's ~280 and very smooth with this new pull request.
- Whereas "time to send request to Redis" was ~1ms in stage without `sentinel_refresh_interval`/`transport_cache_time_in_seconds` and ~0.1ms with it, it's ~0.01ms with this new pull request.
- Simulated Redis master failover via Sentinel was the same across all three configurations: Quick bout of errors while Sentinel picked a new master, and then recovered within just a few seconds.

Given those excellent results, I am landing this PR.

PR #34, opened 2017-12-21
Add stub_action helper for use as decorator or context manager in tests

nickwilliams-eventbrite:


nickwilliams-eventbrite:
@sophieevb:
>Overall looks good to me. Can you add a test for stubbing same action twice in same test with different expected responses, in light of this bug I filed against old SOA: https://jira.evbhome.com/browse/EB-70379

This is already done. Instead of calling `stub_action` multiple times for a given action, we call it once with multiple expectations and multiple return assertions. See `test_two_stubs_same_service_as_decorator_multiple_calls_to_one` and `test_as_decorator_with_patch_after` for tests that prove this behavior.

PR #35, opened 2017-12-22
Improve the msgpack serializer to support local-date and dateless-time objects

nickwilliams-eventbrite:


andrewgodwin:
This was, in fact, intentional - I wanted to keep the surface area of native types we transported to a minimum (as for every one of these types we need to define how they transport into JSON and potentially support them in other languages too). I think I'm OK with DATE, as that's easy to pass around, but what do we have that needs just TIME?

nickwilliams-eventbrite:
>This was, in fact, intentional - I wanted to keep the surface area of native types we transported to a minimum (as for every one of these types we need to define how they transport into JSON and potentially support them in other languages too). I think I'm OK with DATE, as that's easy to pass around, but what do we have that needs just TIME?

I noticed the problem with `DATE` when I was working with switches and accidentally had whole switches (instead of just the values) being transported. Obviously that was incorrect, but it made me review our model types and support for them in `msgpack`. I noticed that we have `DATE` and `TIME` fields in our models in addition to `DATETIME`/`TIMESTAMP`, and we wouldn't be able to transport them without these changes.

These three types are fairly universal and are supported in nearly all languages.

I'd note that our JSON serializer doesn't currently support `DATETIME`, either. If we're serious about supporting the JSON serializer, we already have some work to do in it. 🤔

andrewgodwin:
We should throw away any JSON SOA serializers we have left, it's not possible to pass types through JSON losslessly. Put these in for now, then, let's resolve to check the APIs for consistency on using datetimes vs dates vs times another time.

PR #41, opened 2018-01-09
Recognize either settings variable name non-Django services

nickwilliams-eventbrite:


pysoa/server/server.py
```diff
@@ -372,9 +372,16 @@ def main(cls):
...
-                raise ValueError('Cannot find settings variable in settings module %s' % cmd_options.settings)
+                try:
+                    settings_dict = getattr(settings_module, 'settings')
+                except AttributeError:
+                    raise ValueError(
+                        "Cannot find 'SOA_SERVER_SETTINGS' or 'settings' variable in settings module {}.".format(
+                            cmd_options.settings,
+                        )
+                    )
             settings = cls.settings_class(settings_dict)
```
yanguo-eb: What should take precedence when both `SOA_SERVER_SETTINGS` and `settings` are present?

nickwilliams-eventbrite: `SOA_SERVER_SETTINGS`, because it's more explicit ... `settings` could mean anything, but `SOA_SERVER_SETTINGS` is clearly SOA server settings.

PR #44, opened 2018-01-11
Ensure that service name passed to client is always unicode

nickwilliams-eventbrite:
Everything else (settings, requests, responses, etc.) we enforce to be unicode. We don't enforce this for service names at the top level in the client, but we do at a lower level in the remote Redis transport. This results in hard-to-diagnose bugs that are often missed in testing (which uses a local transport).

### DIFF
--- a/pysoa/test/stub_service.py
+++ b/pysoa/test/stub_service.py
@@ -5,6 +5,7 @@ from functools import wraps
 import re
 
 import mock
+import six
 
 from pysoa.client.client import (
     Client,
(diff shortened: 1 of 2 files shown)


pysoa/test/stub_service.py
```diff
@@ -304,10 +308,18 @@ def __enter__(self):
...
+            assert isinstance(service_name, six.text_type), 'Called service name "{}" must be unicode'.format(
+                service_name,
+            )
+
             requests_to_send_to_mock_client = OrderedDict()
             requests_to_send_to_wrapped_client = []
             for i, action_request in enumerate(actions):
                 request_action = getattr(action_request, 'action', None) or action_request['action']
+                assert isinstance(request_action, six.text_type), 'Called action name "{}" must be unicode'.format(
+                    request_action,
```
yanguo-eb: `request_action` and `action_request` are confusing names. Could one of them be renamed?

nickwilliams-eventbrite: Done.

PR #50, opened 2018-02-02
Start request counter at a random value

andrewgodwin:
This ensures we detect request/response mismatches when two clients
share the same underlying connection (otherwise you could have two
requests with value 0).

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import random
 import time
 import uuid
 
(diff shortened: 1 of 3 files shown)


pysoa/client/client.py
```diff
@@ -49,7 +50,9 @@ def __init__(self, service_name, settings):
                 for m in settings['middleware']
             ]
 
-        self.request_counter = 0
+        # Make sure the request counter starts at a random location to avoid clashing with other clients
+        # sharing the same connection
+        self.request_counter = random.randint(1, 1000000)
```
vartec: I have doubt about the name of the variable, but I can't think of better one right now, so ¯\_(ツ)_/¯

nickwilliams-eventbrite: A fair point, but I can't think of a better name, either.

PR #51, opened 2018-02-02
Removing randint mock

vartec:
Removing randint mock, which causes PR #50 to fail


nickwilliams-eventbrite:
Fascinating. I understand why that was a problem, but I never would have anticipated it.

PR #52, opened 2018-02-02
Add ability to fork multiple server processes with the standalone command

nickwilliams-eventbrite:
The existing standalone command can still be invoked as-is to start a single service process:

```
python3 -m example_service.standalone -s settings_docker
```

However, you can now start multiple service processes, using forking/multiprocessing, but simply adding another option:

```
python3 -m example_service.standalone -s settings_docker --fork 5
```

That command starts 5 service processes.


pysoa/server/standalone.py
```diff
@@ -19,14 +22,63 @@
...
+    )
+    return parser
+
+
+def _get_args(parser):
+    return parser.parse_known_args()[0]
+
+
+def _run_server(args, server_class):
+    if args.fork_processes > 1:
```
yanguo-eb: Do you want to have a max cap? maybe `(multiprocessing.cpu_count() or 1) * 10`

nickwilliams-eventbrite: That's not a bad idea. I'm actually going to set it a bit lower than that—5 processes per CPU. Thanks!

pysoa/server/standalone.py
```diff
@@ -19,14 +22,63 @@
...
+        for i in range(0, args.fork_processes):
+            process = multiprocessing.Process(target=server_class.main, name='pysoa-worker-{}'.format(i))
+
+            if not cmd['keep_starting']:
+                # Got shutdown signal while starting up, so don't keep starting up
+                break
+
+            processes.append(process)
+            process.start()
+            time.sleep(0.3)
```
vartec: why the delay, why check if you should or should not keep starting in the middle of launching?

Why not just 
```

processes = [multiprocessing.Process(target=server_class.main, name='pysoa-worker-{}'.format(i)) for i in range(0, args.fork_processes)]
_ = (p.start() for p in processes) 

```

pysoa/server/standalone.py
```diff
@@ -19,14 +22,63 @@
...
+        import time
+
+        processes = []
+        cmd = {'keep_starting': True}
+
+        def _stop(*_):
+            cmd['keep_starting'] = False
+
+        signal.signal(signal.SIGINT, _stop)
+        signal.signal(signal.SIGTERM, _stop)
```
vartec: you're only using these signals _only_ to stop startup, and you're not removing the handlers after the startup is done.

pysoa/server/standalone.py
```diff
@@ -19,14 +22,63 @@
...
+            if not cmd['keep_starting']:
+                # Got shutdown signal while starting up, so don't keep starting up
+                break
+
+            processes.append(process)
+            process.start()
+            time.sleep(0.3)
+
+        for process in processes:
+            process.join()
```
vartec: But how would the sub-processes know to join, if the signals to terminate are not passed to them?

pysoa/server/standalone.py
nickwilliams-eventbrite: Yes, this was actually quite intentional. In fact, the examples I found online said that the parent process needs to set the signal to `signal.SIG_IGN` or some other function that ignores it, because that can cause the parent process to stop `.join`-ing and it won’t wait before the child processes stop before stopping itself. I did it this way to be able to stop the startup process.

pysoa/server/standalone.py
nickwilliams-eventbrite: I'm going to try this approach and see how well it works.

pysoa/server/standalone.py
nickwilliams-eventbrite: Yep!

pysoa/server/standalone.py
nickwilliams-eventbrite: Turns out that `_ = (p.start() for p in processes) ` is a no-op, because the optimizer optimizes it away. TIL. So had to do actual loop. But otherwise adopted your suggestions.

PR #53, opened 2018-02-03
Fix bug causing response mix-ups with transport cache

nickwilliams-eventbrite:
When the transport cache is enabled, separate clients on separate threads with identical transport settings could get their responses mixed up, such that "client 2" could get responses meant for "client 1" and "client 1" responses meant for "client 2." So, that's bad. This commit fixes that and also adds a test to confirm the problem and a test to confirm the fix.

### DIFF
--- a/pysoa/common/transport/base.py
+++ b/pysoa/common/transport/base.py
@@ -12,7 +12,15 @@ is not business logic. For example, if your implementation has multiple serializ
 types, the metadata may include a mime type to tell the endpoint receiving the message
 which type of serializer to use.
 """
+from __future__ import absolute_import, unicode_literals
+
+import threading
+
 from pysoa.common.metrics import NoOpMetricsRecorder
+
+
+def get_hex_thread_id():
+    return '{:012x}'.format(threading.current_thread().ident)
 
 
 class ClientTransport(object):
--- a/pysoa/common/transport/redis_gateway/client.py
+++ b/pysoa/common/transport/redis_gateway/client.py
@@ -2,7 +2,10 @@ from __future__ import absolute_import, unicode_literals
 
 import uuid
 
-from pysoa.common.transport.base import ClientTransport
+from pysoa.common.transport.base import (
+    ClientTransport,
+    get_hex_thread_id,
+)
 from pysoa.common.transport.redis_gateway.backend.base import BaseRedisClient
 from pysoa.common.transport.redis_gateway.core import RedisTransportCore
 from pysoa.common.transport.redis_gateway.utils import make_redis_queue_name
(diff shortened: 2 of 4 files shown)


pysoa/common/transport/base.py
```diff
@@ -12,9 +12,15 @@
 types, the metadata may include a mime type to tell the endpoint receiving the message
 which type of serializer to use.
 """
+import threading
+
 from pysoa.common.metrics import NoOpMetricsRecorder
 
 
+def get_hex_thread_id():
+    return '{:012x}'.format(hash(threading.current_thread().ident))
```
vartec: Minor nitpick:

I think `hash()` can be skipped. `.ident` is already guaranteed to be a unique integer. 

With high enough integers `hash` can cause clashes,
for example `hash(2305843009213693952) == hash(1)`
The clash is highly improbable, but all it saves is 4 characters. 
So why not `'{:016x}'.format(threading.current_thread().ident)`?

pysoa/common/transport/redis_gateway/client.py
```diff
@@ -29,15 +32,21 @@ def requests_outstanding(self):
 
     def send_request_message(self, request_id, meta, body):
         self._requests_outstanding += 1
-        meta['reply_to'] = self._receive_queue_name
+        meta['reply_to'] = '{receive_queue_name}{thread_id}'.format(
```
vartec: Maybe some separator char between queue name and thread_id.
Let's say `'{receive_queue_name}_{thread_id}'`

tests/common_tests/transport/redis_gateway/test_threading.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def run(self):
+        while self._continue is True:
+            try:
+                # print('  - server receive')
+                request_id, meta, body = self._transport.receive_request_message()
+                # print('  - server send')
+                self._transport.send_response_message(request_id, meta, body)
+            except MessageReceiveTimeout:
+                time.sleep(0.05)
```
vartec: I'm not sure if `freezegun` works fine with threads, but if it does, I'd much rather see `freezegun.tick()` instead of `time.sleep()` in a test. 

OTOH, consider this a non-blocking nitpick.

pysoa/common/transport/base.py
nickwilliams-eventbrite: Good point. Done.

pysoa/common/transport/redis_gateway/client.py
nickwilliams-eventbrite: There's already a separator. The receive queue name is defined above as:

```
        self._receive_queue_name = '{send_queue_name}.{client_id}{response_queue_specifier}'.format(
            send_queue_name=self._send_queue_name,
            client_id=self.client_id,
            response_queue_specifier=BaseRedisClient.RESPONSE_QUEUE_SPECIFIER,
        )
```

And `BaseRedisClient.RESPONSE_QUEUE_SPECIFIER` is an exclamation point.

tests/common_tests/transport/redis_gateway/test_threading.py
nickwilliams-eventbrite: Ehhh ... that seems like more trouble than it's worth. :-)

PR #54, opened 2018-02-05
Add support for auto-reloading code changes in dev environments

nickwilliams-eventbrite:
Using the `--use-file-watcher` option when starting a service with `example_service` or `python -m example_service.standalone` will now result in the service's being wrapped by a reloader that watches for file changes and restarts the service when file changes are detected. This feature is extremely useful for, and also only intended for, development environments. It is not recommended for production environments. It is fully compatible with both standard mode and forking mode (`--fork-processes` / `--fork` / `-f`).


pysoa/server/standalone.py
```diff
@@ -59,8 +70,16 @@ def _run_server(args, server_class):
...
+        processes = []
+
+        def _sigterm_forks(*_, **__):
+            for process in processes:
+                process.terminate()
+
+        # We don't want these signals to actually kill this process; just sub-processes
         signal.signal(signal.SIGINT, signal.SIG_IGN)
         signal.signal(signal.SIGTERM, signal.SIG_IGN)
+        signal.signal(signal.SIGUSR1, _sigterm_forks)  # special signal by reloader says we actually need to propagate
```
vartec: Why not `SIGHUP`?

PR #56, opened 2018-02-06
Refactor expansion

yanguo-eb:
- Refactor out expansion method in client
- renaming to differentiate expansion_config init v.s. expansion_query from request
- When make request, the `body` takes `[value]` instead of `value`, assuming we always call batch endpoints
- When expand, the initial `exp_service_requests` set to empty, because the upstream `service` has been called before this method.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -158,7 +158,7 @@ class Client(object):
     settings_class = PolymorphicClientSettings
     handler_class = ServiceHandler
 
-    def __init__(self, config, expansions=None, settings_class=None, context=None):
+    def __init__(self, config, expansion_config=None, settings_class=None, context=None):
         """
         Args:
             config: dict of {service_name: service_settings}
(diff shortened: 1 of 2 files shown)


pysoa/client/client.py
```diff
@@ -254,7 +254,7 @@ def call_action(
             **kwargs
         ).actions[0]
 
-    def call_actions(self, service_name, actions, expansions=None, raise_action_errors=True, **kwargs):
+    def call_actions(self, service_name, actions, expansion_query=None, raise_action_errors=True, **kwargs):
```
andrewgodwin: I would probably leave this called `expansions` for consistency with our old stuff, but I agree with the other change to `expansion_config` so they're still named differently.

nickwilliams-eventbrite: I agree with this comment 100%. Otherwise this looks great.

yanguo-eb: Agree. Updated.

PR #61, opened 2018-02-12
Add support for server introspection

nickwilliams-eventbrite:
This commit creates a new stock introspection action, which returns information about the service and its actions and their request and response schemas, and uses it by default if the server does not define an `introspect` action. As a related feature, it also ensures that every server responds to `status` if the server does not define a `status` action. Finally, it ensures the status action is adequately documented, so that it can be introspected well.

### DIFF
--- a/pysoa/server/action/introspection.py
+++ b/pysoa/server/action/introspection.py
@@ -1,0 +1,160 @@
+from __future__ import absolute_import, unicode_literals
+
+from conformity import fields
+import six
+
+from pysoa.common.constants import ERROR_CODE_INVALID
+from pysoa.common.types import Error
+from pysoa.server.action.base import Action
+from pysoa.server.action.status import BaseStatusAction
+from pysoa.server.errors import ActionError
+from pysoa.server.server import Server
+
+
+class IntrospectionAction(Action):
+    """
+    This action returns detailed information about the service's defined actions and the request and response schemas
+    for each action, along with any documentation defined for the action or for the service itself. It can be passed
+    a single action name to return information limited to that single action. Otherwise, it will return information for
+    all of the service's actions.
+
+    This action will be added to your service on your behalf if you do not define an action with name `introspect`.
+
+    Making your services and actions capable of being introspected is simple. If your server class has a `description`
+    attribute, that will be the service's documentation that introspection returns. If your server class does not have
+    this attribute but does have a docstring, introspection will use the docstring. The same rule applies to action
+    classes: Introspection first looks for a `description` attribute and then uses the docstring, if any. If neither of
+    these are found, the applicable service or action documentation will be done.
+
+    Introspection then looks at the `request_schema` and `response_schema` attributes for each of your actions, and
+    includes the details about these schemas in the returned information for each action. Be sure you include field
+    descriptions in your schema for the most effective documentation possible.
+    """
+    description = (
+        "This action returns detailed information about the service's defined actions and the request and response "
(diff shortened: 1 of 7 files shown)


pysoa/server/action/introspection.py
```diff
@@ -0,0 +1,148 @@
...
+                description='A dict mapping action names to action description dictionaries. This contains details '
+                            'about every action in the service unless `action_name` is specified in the request body, '
+                            'in which case it contains details only for that action.',
+            ),
+        },
+        optional_keys=('documentation', ),
+    )
+
+    def __init__(self, server):
+        super(IntrospectionAction, self).__init__(server.settings)
```
bar: `.__init__(server.settings)` feels wrong

nickwilliams-eventbrite: Care to elaborate? I don't see what the issue is. Can't read your mind 😄.

bar: sorry, I meant that `IntrospectionAction` inherits from `Action` who's signature is
```
    def __init__(self, settings=None):
        self.settings = settings
```
meaning we should make it clear that the first argument is not `settings` you'd expect when instantiating the action
```
def __init__(self, server, settings=None):
    super(IntrospectionAction, self).__init__(settings=server.settings)
```

nickwilliams-eventbrite: Okay, I understand. You're concerned that the conflicting interfaces are confusing. I'll resolve that with documentation.

pysoa/server/action/introspection.py
```diff
@@ -0,0 +1,155 @@
...
+        "This action returns detailed information about the service's defined actions and the request and response "
+        "schemas for each action, along with any documentation defined for the action or for the service itself. It "
+        "can be passed a single action name to return information limited to that single action. Otherwise, it will "
+        "return information for all of the service's actions."
+    )
+
+    request_schema = fields.Dictionary(
+        {
+            'action_name': fields.UnicodeString(
+                min_length=2,
```
vartec: Do we disallow single letter actions?

pysoa/server/action/introspection.py
```diff
@@ -0,0 +1,155 @@
...
+                ),
+                description='A dict mapping action names to action description dictionaries. This contains details '
+                            'about every action in the service unless `action_name` is specified in the request body, '
+                            'in which case it contains details only for that action.',
+            ),
+        },
+        optional_keys=('documentation', ),
+    )
+
+    def __init__(self, server):
```
vartec: It feels a bit weird, to have different params in subclass. Workaround could be to
```
def __init__(self, settings=None, server=None)
   if server is None:
       raise TypeError('server parameter is required')
  super(IntrospectionAction, self).__init__(settings or server.settings)
```
and calling side
`IntrospectionAction(server=server)`

pysoa/server/action/introspection.py
nickwilliams-eventbrite: No. I don't suppose we do. Fixed.

pysoa/server/action/introspection.py
nickwilliams-eventbrite: I don't understand the resistance I'm getting to this. I've never heard of this being improper, and it's certainly not causing any problems. The suggested alterations don't actually improve anything. 🤔

nickwilliams-eventbrite:
@bar and @vartec: I applied a hybrid approach of your respective suggestions. Hopefully it meets your expectations.

PR #62, opened 2018-02-13
Fix unit test failures caused by Django connection cleanup

nickwilliams-eventbrite:
When unit tests are run, the Django test fixture transaction wrapping can break the PySOA server connection cleanup, manifesting as an strange atomic block error on the first database query within an action. This commit ensures we do not clean up connections if we're in the middle of a transaction, which solves that problem. It also ensures that connection cleanup failures caused by PyTest's `mark.django_db` test are ignored.

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -300,9 +300,20 @@ class Server(object):
 
     def _close_old_django_connections(self):
         if self.use_django:
-            from django.db import close_old_connections
-            self.logger.debug('Cleaning Django connections')
-            close_old_connections()
+            try:
+                from django.db import transaction
+                if transaction.get_autocommit():
+                    from django.db import close_old_connections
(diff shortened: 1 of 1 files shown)


pysoa/server/server.py
```diff
@@ -300,9 +300,18 @@ def setup(self):
...
-            from django.db import close_old_connections
-            self.logger.debug('Cleaning Django connections')
-            close_old_connections()
+            try:
+                from django.db import transaction
+                if transaction.get_autocommit():
+                    from django.db import close_old_connections
+                    self.logger.debug('Cleaning Django connections')
+                    close_old_connections()
+            except BaseException as e:
```
vartec: why `BaseException` and not `Exception`?

nickwilliams-eventbrite: Because PyTest's `Failed` exception extends `BaseException`, not `Exception`.

PR #63, opened 2018-02-14
Add request details to a logging context for all log records

nickwilliams-eventbrite:


pysoa/server/logging.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class PySOALogContextFilter(logging.Filter):
+    def __init__(self):
+        super(PySOALogContextFilter, self).__init__('')
+
+    def filter(self, record):
+        context = self.get_logging_request_context()
+        if context:
+            record.correlation_id = context.get('correlation_id') or ''
```
adamsussman: Not that I disagree in any way with this, but I had thought that correlation_id was an ebsoa thing rather than a pysoa thing, although I would argue that it ought to be a base level pysoa thing anyway.

nickwilliams-eventbrite: It's not just n EB SOA thing. It's base PySOA: https://github.com/eventbrite/pysoa/blob/73e8746f7aafd969b0649b9175daf2af77b59f7d/pysoa/client/client.py#L402-L424

tests/server/test_logging.py
```diff
@@ -0,0 +1,94 @@
...
+            PySOALogContextFilter.get_logging_request_context()
+        )
+
+        self.assertIsNone(thread_data['first_get'])
+        self.assertEqual({'foo': 'bar', 'baz': 'qux'}, thread_data['second_get'])
+        self.assertEqual({'foo': 'bar', 'baz': 'qux'}, thread_data['third_get'])
+
+        thread_data['do_clear'] = True
+
+        thread = threading.Thread(target=fn)
```
vartec: I don't like idea of spawning real threads in tests, but I guess as long as you don't put any asserts in there, it ought to be fine...

pysoa/server/settings.py
```diff
@@ -59,7 +59,13 @@ class ServerSettings(SOASettings):
                 ),
                 'filters': fields.SchemalessDictionary(
                     key_type=fields.UnicodeString(),
-                    value_type=fields.Dictionary({'name': fields.UnicodeString()}, optional_keys=('name', )),
+                    value_type=fields.Dictionary(
+                        {
+                            '()': fields.Anything(description='The optional filter class'),
```
vartec: why `'()'` as a field name?

tests/server/test_logging.py
nickwilliams-eventbrite: No other real way to test that it works properly under threads. :-)

pysoa/server/settings.py
nickwilliams-eventbrite: Because that's the way Python Logging config works. To create a logging filter of a specific class, you use `'()'` as the field name. I have no idea why. It's ugly as sin. But it's part of Python; not in our control.

PR #64, opened 2018-02-14
Fix expansion response format

yanguo-eb:
Currently there are 5 issues with expansion, and this PR addresses the first 2.
    Issues (current --- proposed fix)
    1. source field is erased in response after expansion. --- source field should be kept untouched in response
    2. expansion expect a singular endpoint, e.g. `get_xx_by_id`. --- expansion should expect a plural endpoint, e.g. `get_xxs_by_ids`, with additional layer of id-key.
    3. expansion dispatch request one by one --- expansion should group requets by endpoint and do them in batch
    4. expansion has no permission check --- TBD
    5. expansion has no proper error handling or error code returned to the client --- TBD

    The first 2 issue are easy to fix, as in thie PR
    The third requires a overhaul, and will be in a separate PR
    The 4th and 5th requires more discussion.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -362,10 +362,10 @@ class Client(object):
                             # If everything is okay, replace the expansion object with the response value
                             if exp_action_response.body:
                                 value = exp_action_response.body[expansion_node.response_field]
-                                # Add the expansion value to the object and remove the
-                                # source field.
-                                del exp_object[expansion_node.source_field]
-                                exp_object[expansion_node.dest_field] = value
+                                # Add the expansion value to the object
+                                # Assume there is one item, and discard the id-key
+                                (dest_obj, ) = value.values()
+                                exp_object[expansion_node.dest_field] = dest_obj
 
                             # Potentially add additional pending expansion requests.
                             if expansion_node.expansions:
--- a/pysoa/client/expander.py
+++ b/pysoa/client/expander.py
@@ -1,5 +1,7 @@
 import six
 
+from pysoa.common.types import Error
+
 
 class TypeNode(object):
     """
@@ -282,6 +284,12 @@ class ExpansionConverter(object):
                     if not child_expansion_node:
                         type_expansion = self.type_expansions[expansion_node.type][exp_name]
                         type_route = self.type_routes[type_expansion['type']]
+                        if type_expansion['dest_field'] == type_expansion['source_field']:
+                            raise Error(
+                                'Expansion configuration dest_field error: '
+                                'dest_field can not have the same name as the source_field: '
+                                '{}'.format(type_expansion['source_field'])
+                            )
                         child_expansion_node = ExpansionNode(
                             type=type_expansion['type'],
                             name=exp_name,
--- a/tests/client/test_expansions.py
+++ b/tests/client/test_expansions.py
@@ -12,27 +12,27 @@ class TestClientWithExpansions(TestCase):
             'type_routes': {
                 'author_router': {
                     'service': 'author_info_service',
-                    'action': 'get_author_by_id',
-                    'request_field': 'id',
-                    'response_field': 'author_detail',
+                    'action': 'get_authors_by_ids',
+                    'request_field': 'ids',
+                    'response_field': 'authors_detail',
                 },
                 'publisher_type': {  # Note: Likely a bug here
                     'service': 'publisher_info_service',
-                    'action': 'get_publisher_by_id',
-                    'request_field': 'id',
-                    'response_field': 'publisher_detail',
+                    'action': 'get_publishers_by_ids',
+                    'request_field': 'ids',
+                    'response_field': 'publishers_detail',
                 },
                 'address_router': {
                     'service': 'address_info_service',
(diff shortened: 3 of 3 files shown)


pysoa/client/client.py
```diff
@@ -364,9 +364,15 @@ def _expand_objects(self, objs_to_expand):
...
                                 # source field.
-                                del exp_object[expansion_node.source_field]
-                                exp_object[expansion_node.dest_field] = value
-
+                                if expansion_node.discard_source_field:
+                                    del exp_object[expansion_node.source_field]
+                                if expansion_node.discard_dest_field_id_key:
+                                    # value is a dict with one item. We want to throw away the id key,
+                                    # only put the object in place
+                                    (_, dest_obj), = value.items()
```
vartec: dest_obj  = value.values()[0]

yanguo-eb: hmmm
TypeError: 'dict_values' object does not support indexing

nickwilliams-eventbrite: You shouldn't use `items()` here, because it returns something different in Python 3 than it does in Python 2.

vartec: ah... right, in Python 3 `items()` and `values()` are iterators, not lists... 
Still you can do `dest_obj, = value.values()` which which will work in both.

nickwilliams-eventbrite: Yep, that should work fine, as long as this for sure only _**ever**_ has one value. If it could possibly have more than one, you'd get `ValueError: too many values to unpack`.

tominsam: I believe this is a misunderstanding of how expansions are supposed to work.

what's supposed to happen is that you call the SOA action once per _type_, and you pass an array of all the IDs of all the objects of that type. The action returns a dict mapping those IDs to the expansion responses, and the client is supposed to re-thread the dict values onto the object.

The point of expansions is that if I have a list response of 50 events, and I expand, say, "format" on those events, the system makes a single SOA call and passes the set of _distinct_ format_ids of the events (so if the events have all the same format it's only inflating one obejct), and the SOA action responds with a dict of format_id -> format, then the server applies those format objects to the events.

The pysoa implementation here seems to be making an SOA call per object in the response, which is why the expansion action response here has a single value.

pysoa/client/expander.py
```diff
@@ -282,6 +284,12 @@ def dict_to_trees(self, exp_dict):
                     if not child_expansion_node:
                         type_expansion = self.type_expansions[expansion_node.type][exp_name]
                         type_route = self.type_routes[type_expansion['type']]
+                        if type_expansion['dest_field'] == type_expansion['source_field']:
+                            raise Error(
+                                "Expansion configuration dest_field error: "
+                                "dest_field can not have the same name as the source_field: "
+                                "{}".format(type_expansion['source_field'])
```
nickwilliams-eventbrite: Single quotes, please.

pysoa/client/client.py
```diff
@@ -362,10 +362,10 @@ def _expand_objects(self, objs_to_expand):
                             # If everything is okay, replace the expansion object with the response value
                             if exp_action_response.body:
                                 value = exp_action_response.body[expansion_node.response_field]
-                                # Add the expansion value to the object and remove the
-                                # source field.
-                                del exp_object[expansion_node.source_field]
-                                exp_object[expansion_node.dest_field] = value
+                                # Add the expansion value to the object
+                                # Assume there is one item, and discard the id-key
+                                (dest_obj, ) = value.values()
```
andrewgodwin: Use `values()[0]` here rather than pattern matching.

yanguo-eb:
```TypeError: 'dict_values' object does not support indexing
```

pysoa/client/expander.py
yanguo-eb: right. will edit.

pysoa/client/client.py
nickwilliams-eventbrite: Yeah, `value.values()[0]` is Python 2-specific. Perhaps `list(value.values())[0]`, but, at that point, is that really the best way to do it? 🤔

nickwilliams-eventbrite: As long as we are _sure_ that there is only ever one value in this dict, then I would do it exactly like Yan has it. Performance is considerably better: 

```
In [13]: %timeit (dest_obj, ) = d.values()
151 ns ± 2.57 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [15]: %timeit dest_obj = list(d.values())[0]
352 ns ± 10.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

However, if there's any chance at all that this dict has more than one value (which I would argue calls into question this code in general), then it needs to be `list(value.values())[0]`.

yanguo-eb: It is an invariant that there is one and only one item returned in `value`, because the request to `get_xxxs_by_ids` dispatches with a one-element array, thus the returned response should be a one-element collection too. Otherwise, there is a data inconsistency and it should fail explicitly with `ValueError` during unpacking.

PR #67, opened 2018-02-16
[PATCH] If no databases are configured, do not attempt Django connection cleanup

nickwilliams-eventbrite:
If no `DATABASES` aliases are configured in the Django settings, attempting connection cleanup causes errors. So don't attempt said cleanup in this case.

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -314,8 +314,13 @@ class Server(object):
 
     def _close_old_django_connections(self):
         if self.use_django:
+            from django.conf import settings
+            if not getattr(settings, 'DATABASES'):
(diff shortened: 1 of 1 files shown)


pysoa/server/server.py
```diff
@@ -314,8 +314,13 @@ def setup(self):
 
     def _close_old_django_connections(self):
         if self.use_django:
+            from django.conf import settings
+            if not getattr(settings, 'DATABASES'):
+                # No database connections are configured, so we have nothing to do
+                return
+
+            from django.db import transaction
```
vartec: nice catch pulling that out of the `try`

PR #74, opened 2018-02-28
[MINOR] Add support for metric timer resolution

nickwilliams-eventbrite:
Currently, timer metrics default to millisecond resolution. With this change, timer metrics will now also support microsecond and nanosecond resolution.

### DIFF
--- a/pysoa/common/metrics.py
+++ b/pysoa/common/metrics.py
@@ -1,5 +1,7 @@
 from __future__ import unicode_literals
 
+import enum
+
 from conformity import fields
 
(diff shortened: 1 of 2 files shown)


nickwilliams-eventbrite:
>Why import a whole enum third-party package rather than just having a normal class with attributes?

@andrewgodwin, `enum` is a builtin in Python 3 with a backport for Python 2. An enum is a better semantical representation of what this is. Your point, however, makes me realize that this should be a Python 2-only requirement.

PR #76, opened 2018-03-13
suppress expansion exception when request has invalid key

yanguo-eb:
When doing expansion, a key exception should be silenced because it is a client error not a server error. Consequently return only the original skeleton_response.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -188,6 +188,9 @@ class Client(object):
     class ImproperlyConfigured(Exception):
         pass
 
+    class InvalidExpansionKey(Exception):
+        pass
+
     class JobError(Exception):
         """
         Raised by Client.call_action(s) when a job response contains one or more job errors.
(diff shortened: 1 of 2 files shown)


pysoa/client/client.py
```diff
@@ -297,8 +301,12 @@ def call_actions(self, service_name, actions, expansions=None, raise_action_erro
     def _perform_expansion(self, skeleton_response, expansions):
         # Perform expansions
         if expansions and hasattr(self, 'expansion_converter'):
-            objs_to_expand = self._extract_candidate_objects(skeleton_response.actions, expansions)
-            self._expand_objects(objs_to_expand)
+            try:
+                objs_to_expand = self._extract_candidate_objects(skeleton_response.actions, expansions)
+            except KeyError as keyError:
```
vartec: ugh... camelCase? :-P
Just call it `error`

yanguo-eb: thank you :)

PR #77, opened 2018-03-14
[MINOR] Censor sensitive fields in the request and response log

nickwilliams-eventbrite:


pysoa/server/logging.py
```diff
@@ -36,3 +38,151 @@ def get_logging_request_context(cls):
...
+    # 10000000 loops, best of 3: 131 ns per loop
+    #
+    # In [3]: %timeit random_string()
+    # 100000 loops, best of 3: 10.4 us per loop
+    #
+    # In [4]: %timeit random_string() in SENSITIVE_FIELDS
+    # 100000 loops, best of 3: 10.7 us per loop
+    #
+    # Suffice it to say, a lookup against this set should always take less than 250 ns.
+    SENSITIVE_FIELDS = frozenset({
```
adamsussman: I would add the auth token to this list.  Encrypted body content in general ought to be blanked.

adamsussman: How does a given service customize this list for itself?

nickwilliams-eventbrite: We discussed that in the squad channel on Slack. A request was made to //temporarily// exclude auth tokens from this while we work work out the kinks in our new internal authentication service, and then eventually add it to this list.

pysoa/server/logging.py
```diff
@@ -36,3 +38,151 @@ def get_logging_request_context(cls):
...
+    def __bytes__(self):
+        # If this method is called, we must be in Python 3, which means __str__ must be returning a string
+        return self.__str__().encode('utf-8')
+
+    def __unicode__(self):
+        # If this method is called, we must be in Python 2, which means __str__ must be returning bytes
+        return self.__str__().decode('utf-8')
+
+    @classmethod
+    def _copy_and_censor_unknown_value(cls, v, should_censor_strings):
```
vartec: at the point where it's used, can the value be any custom type?

nickwilliams-eventbrite: I'm not sure what you mean by this. Can you re-phrase?

vartec: can `v` be anything but the basic types?

pysoa/server/logging.py
andrewgodwin: I'm not sure that will work forever, but this is a decent enough start.

pysoa/server/logging.py
```diff
@@ -36,3 +38,151 @@ def get_logging_request_context(cls):
...
+        if isinstance(v, dict):
+            return cls._copy_and_censor_dict(v)
+
+        if isinstance(v, (list, set, tuple)):
```
andrewgodwin: I'd add `frozenset` to this list.

pysoa/server/logging.py
nickwilliams-eventbrite: Yes. If it's not one of the basic types, it won't be censored, but we only support serialization/deserialization of the basic types, anyway, so this isn't a problem.

pysoa/server/logging.py
nickwilliams-eventbrite: Oh, good catch. I had thought `frozenset` was-a `set`. Turns out it's not.

PR #79, opened 2018-04-02
[MAJOR] Add support for setting a custom timeout when sending a request

nickwilliams-eventbrite:
This adds support for setting a custom timeout when sending a request, so that users can shorten or lengthen the time the client will block waiting for a response. As part of this major refactor, we add proper `ABCMeta` to several class hierarchies.


cubes:
I looked over the review, and it looks reasonable, but I don’t consider myself familiar enough with pysoa to give it a :ship:.

PR #80, opened 2018-04-05
Add stock ability to include other services' status in status

nickwilliams-eventbrite:
This change adds the standard ability to include other services' `status` action response in the diagnostic information of the `status` action response of a service configured to call those other services. It also includes a small enhancement to `stub_action` necessary to fully test this change to the `BaseStatusAction`. Finally, it includes an enhancement to `BaseStatusAction` to pass the `request` object to `check_` methods.

### DIFF
--- a/pysoa/server/action/status.py
+++ b/pysoa/server/action/status.py
@@ -9,6 +9,12 @@ from conformity import fields
 import six
 
 import pysoa
+from pysoa.common.transport.exceptions import (
+    MessageReceiveError,
+    MessageReceiveTimeout,
+    MessageSendError,
+    MessageSendTimeout,
+)
 from pysoa.server.action import Action
 
 
@@ -21,12 +27,12 @@ class BaseStatusAction(Action):
     If you want to use default StatusAction use StatusActionFactory(version), passing in the version of your service.
     If you do not specify an action with name `status` in your server, this will be done on your behalf.
 
-    If you want to make a custom StatusAction, subclass this class, make self._version return your service's version,
-    self._build optionally return your service's build string, and add any additional health check methods you desire.
-    Health check methods must start with the word check_.
+    If you want to make a custom StatusAction, subclass this class, make self._version return your service's version
+    string, self._build optionally return your service's build string, and add any additional health check methods you
+    desire. Health check methods must start with the word check_.
 
-    Health check methods must take no arguments, and return a list of tuples in the format
-    `(is_error, code, description)`.
+    Health check methods accept a single argument, the request object (an instance of `ActionRequest`), and return a
+    list of tuples in the format `(is_error, code, description)` (or a false-y value if there are no problems):
 
         - is_error: `True` if this is an error, `False` if it is a warning.
         - code: Invariant string for this error, like "MYSQL_FAILURE"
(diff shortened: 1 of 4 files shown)


pysoa/server/action/status.py
```diff
@@ -147,6 +168,44 @@ def run(self, request):
 
         return status
 
+    def _check_client_settings(self, request):
+        """
+        This method checks any client settings configured for this service to call other services, calls the `status`
+        action of each configured service with `verbose: False`, adds that diagnostic information, and reports any
```
vartec: could you add here that `verbose: False` guarantees no further recursive calls?

nickwilliams-eventbrite: Good thinking. Done.

pysoa/server/action/status.py
```diff
@@ -147,6 +169,44 @@ def run(self, request):
...
+            # There's no need to even add diagnostic details if no client settings are configured
+            return
+
+        self.diagnostics['services'] = {}
+
+        problems = []
+
+        for service_name, _ in six.iteritems(request.client.settings):
+            try:
+                response = request.client.call_action(service_name, 'status', body={'verbose': False})
```
cubes: There's a subtle bug here, which is that if the remote service is down, this will wait the default configured timeout before raising a MessageSendTimeout or MessageReceiveTimeout exception. The problem is that we tend to set the same default timeout everywhere, so the caller of the healthcheck will timeout too.

Thus, you need to specify a lower timeout here using the new per-request timeout functionality you recently added. I would also suggest that rather than calling each remote service in serial, you use the `call_actions` to call all the remote services in parallel with the same timeout, which I suggest be half of the default timeout things are normally configured with.

Thoughts?

pysoa/server/action/status.py
```diff
@@ -147,6 +169,44 @@ def run(self, request):
...
+        This method checks any client settings configured for this service to call other services, calls the `status`
+        action of each configured service with `verbose: False` (which guarantees no further recursive status checking),
+        adds that diagnostic information, and reports any problems. To include this check in your status action, define
+        `check_client_settings = BaseStatusAction._check_client_settings` in your status action class definition.
+        """
+        if not request.client.settings:
+            # There's no need to even add diagnostic details if no client settings are configured
+            return
+
+        self.diagnostics['services'] = {}
```
cubes: Nitpick, I called the top-level key that I nested pysoa service diagnostics `pysoa_services`. I don't feel strongly about the name, but Eventbrite ops will need to adjust if we change it.

I have a slight preference for the more specific `pysoa_services` vs. `services` to support the use case where pysoa is used together with other service frameworks.

nickwilliams-eventbrite: I have used `pysoa_services` in the health checks for legacy SOA services (such as Decision Service), to intentionally indicate that they are non-legacy-SOA services. However, as this is a built-in specifically for PySOA services only (and as it's part of our public OSS software), I think `services` fits better here.

pysoa/server/action/status.py
nickwilliams-eventbrite: All good points.

1) As suggested, I have supplied a lower timeout. It is now 2 seconds. Don't really want to set it any lower.

2) Good idea that this should run in parallel. Unfortunately, the `Client` doesn't yet actually support parallel calling. `call_actions` only supports one service, and it's serial. I'm working on changes now to support parallel calls, but it's not done yet. So, this will have to suffice for now. I will circle back at a later time to improve this.

pysoa/server/action/status.py
cubes: 👍

PR #83, opened 2018-04-11
[MINOR] Ensure stub_action supports expansions

nickwilliams-eventbrite:
The `stub_action` tool was wrapping `Client.call_actions` and `Client.call_jobs_parallel`, which resulted in expansions being completely ignored and not processed. This, itself, caused several problems, but it also meant that `Client.send_request` and `Client.get_all_responses` bypassed `stub_action` completely. This latest—and hopefully final—iteration of `stub_action` leaves `call_actions` and `call_jobs_parallel` alone and, instead, wraps `send_request` and `get_all_responses`. The result is that all routes to making a request pass through the stubbing code (when activated), and expansions are properly processed.

### DIFF
--- a/pysoa/test/stub_service.py
+++ b/pysoa/test/stub_service.py
@@ -1,6 +1,9 @@
 from __future__ import absolute_import, unicode_literals
 
-from collections import OrderedDict
+from collections import (
+    defaultdict,
+    OrderedDict,
+)
 from functools import wraps
 import re
 
@@ -14,13 +17,8 @@ from pysoa.client.client import (
 from pysoa.client.settings import ClientSettings
 from pysoa.common.metrics import NoOpMetricsRecorder
 from pysoa.common.transport.exceptions import (
-    ConnectionError,
-    InvalidMessageError,
     MessageReceiveError,
     MessageReceiveTimeout,
-    MessageSendError,
-    MessageSendTimeout,
-    MessageTooLarge,
 )
 from pysoa.common.transport.local import LocalClientTransport
 from pysoa.common.types import (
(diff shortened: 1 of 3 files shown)


pysoa/test/stub_service.py
```diff
@@ -301,165 +312,178 @@ def __init__(self, service, action, body=None, errors=None):
...
-        @wraps(Client.call_jobs_parallel)
-        def wrapped_call_jobs_parallel(client, jobs, *args, **kwargs):
-            jobs_to_reinsert = set()
-            actions_to_reinsert = OrderedDict()
-            for j_i, job_request in enumerate(jobs):
-                service_name, actions = job_request['service_name'], job_request['actions']
-                assert isinstance(service_name, six.text_type), 'Called service name "{}" must be unicode'.format(
+            actions_to_send_to_mock = OrderedDict()
+            actions_to_send_to_wrapped_client = []
+            for i, action_request in enumerate(actions):
```
yanguo-eb: nitt: single letter var name.

pysoa/test/stub_service.py
```diff
@@ -184,6 +182,19 @@ def stub_action(self, action, body=None, errors=None):
         self.action_class_map[action] = _make_stub_action(action, body, errors)
 
 
+class _StubActionRequestCounter(object):
+    def __init__(self):
+        self._counter = 0
+
+    def get_next(self):
+        value = self._counter
+        self._counter += 1
```
yanguo-eb: Do we need to consider thread safety here?

pysoa/test/stub_service.py
nickwilliams-eventbrite: My understanding is that this is considered okay when the single-letter is used for the enumerated index of an iterating object.

pysoa/test/stub_service.py
nickwilliams-eventbrite: No, because this is used exclusively for unit tests, which are run single-threaded.

PR #84, opened 2018-04-12
[MAJOR] Fix several expansion bugs and refactor configuration

nickwilliams-eventbrite:
This change fixes the following several bugs in expansions:

- Expansions fail with `KeyError` if only some of the base objects contain the source field or action errors if the source field in one of the base objects is `None`.
- Expansions call expansion actions multiple times (once for every expansion object) instead of one time (with values from every expansion object). Fixing this required a major overhaul of the code that performs expansions.
- Expansions do not pass the correlation ID, context overrides, and control information from the parent request to the expansion requests, which can omit critical authentication information.
- Expansions do not properly process nested expansions unless the nested types and their routes have the same name, which can be extremely confusing. Fixing this required breaking changes to the Type Expansions Configuration format, so I took the opportunity to also rename an abbreviated field in said format (`dest_field` -> `destination_field`).

This change also adds an extensive and complex set of tests against a complicated expansion hierarchy, including many nested levels with values including and not including expansion criteria.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import collections
 import random
 import time
 import uuid
@@ -352,8 +353,10 @@ class Client(object):
             error_actions = [action for action in response.actions if action.errors]
             if error_actions:
                 raise self.CallActionError(error_actions)
+
         if expansions:
-            self._perform_expansion(response.actions, expansions)
+            kwargs.pop('continue_on_error', None)
+            self._perform_expansion(response.actions, expansions, **kwargs)
 
         return response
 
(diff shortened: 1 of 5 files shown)


pysoa/client/client.py
```diff
@@ -539,96 +542,112 @@ def call_jobs_parallel(
...
                 )
         return objects_to_expand
 
-    def _expand_objects(self, objects_to_expand):
-            # Keep track of expansion action errors that need to be raised
-            expansion_errors = []
-            # Initialize service request cache
-            exp_service_requests = {}
-            # Loop until we have no outstanding requests or responses
-            while objects_to_expand or any(exp_service_requests.values()):
```
sethbrite: I see you removed this extra check for outstanding requests that have not received a response yet. How does your code handle this?

nickwilliams-eventbrite: That's not a thing that can happen here. `get_all_responses` does just that. It gets _all_ outstanding responses, or it raises a `MessageReceiveTimeout` waiting for any further outstanding responses. `get_all_responses` cannot return only some outstanding requests without raising an exception. So the check for outstanding requests here was unnecessary.

PR #85, opened 2018-04-13
Better handling of out-of-order responses (WIP)

sethbrite:
Added `get_response_for_request` helper that blocks until an expected request is received (or times out). This solves bugs that result from `call_actions` receiving an OOO response via `get_all_responses`.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -34,7 +34,6 @@ class ServiceHandler(object):
     _transport_cache = {}
 
     def __init__(self, service_name, settings):
-        self.service_name = service_name
         self.metrics = settings['metrics']['object'](**settings['metrics'].get('kwargs', {}))
 
         transport_cache_time_in_seconds = 0
--- a/pysoa/test/stub_service.py
+++ b/pysoa/test/stub_service.py
@@ -314,7 +314,6 @@ class stub_action(object):  # noqa
 
         self._stub_action_responses_outstanding = defaultdict(dict)
         self._stub_action_responses_to_merge = defaultdict(dict)
-        self._stub_action_unwrapped_call_id_map = {}
 
     def __enter__(self):
         self._wrapped_client_send_request = Client.send_request
@@ -371,7 +370,6 @@ class stub_action(object):  # noqa
                     request_id,
                     continue_on_error,
                 )
-                self._stub_action_unwrapped_call_id_map[request_id] = unwrapped_request_id
 
             ordered_actions_for_merging = OrderedDict()
             job_response_transport_exception = None


pysoa/client/client.py
```diff
@@ -134,6 +137,9 @@ def send_request(self, job_request, message_expiry_in_seconds=None):
             self.metrics.commit()
 
     def _get_response(self, receive_timeout_in_seconds=None):
+        if self.unexpected_responses:
+            return self.unexpected_responses.popitem(last=False)
```
vartec: if there is an unexpected_response, that never gets consumed, then this is going to keep on popping the same one..

pysoa/client/client.py
```diff
@@ -169,6 +180,45 @@ def get_all_responses(self, receive_timeout_in_seconds=None):
...
+            while start_time < end_time:
+                receive_timeout_in_seconds = int(end_time - start_time) if receive_timeout_in_seconds is not None else None
+                for request_id, response in self.get_all_responses(receive_timeout_in_seconds):
+                    if request_id != expected_request_id:
+                        unexpected_responses[request_id] = response
+                    else:
+                        return response
+                start_time = time.time()
+        finally:
+            self.unexpected_responses = unexpected_responses
```
vartec: Not sure if I understand what it does correctly. 
Let's say you you send jobs `1`, `2` & `3`. The first two are slow, the last one is fast. 
So let's say servers responds in order `3`, `2`, `1`.

Client waiting for job `1` will get responses for jobs `3` and `2`, put it on local `unexpected_responses` and not share to `self.unexpected_responses` until either response for job `1` or timeout comes. Why isn't line 212 directly `self.unexpected_responses`?

nickwilliams-eventbrite:
I have concerns about this solution. I feel like it's adding unnecessary complexity to both `ServiceHandler`, `Client`, and `stub_action` to solve what's otherwise a very simple problem. I'd like to propose an alternative solution:

```python
        responses = list(self.get_all_responses(service_name, receive_timeout_in_seconds=timeout))
-       response_id, response = responses[0]
-       if response_id != request_id:
-           raise Exception('Got response with ID {} for request with ID {}'.format(response_id, request_id))
+       found = False
+       for response_id, response in responses:
+           if response_id == request_id:
+               found = True
+               break
+       if not found:
+           # This error should be impossible if `get_all_responses` is behaving correctly, but let's raise a meaningful error just in case
+           raise Exception('Got unexpected response(s) with ID(s) {} for request with ID {}'.format([r[0] for r in responses], request_id))
```

In justifying this more-simplistic solution, a little explanation is in order. `get_all_responses` can do only one of exactly two things (and no more):

- Return _all_ outstanding responses, and one of those _will_ be the response to this request (even if some previously-err'd responses show up first, which is the problem we're trying to solve)
- Raise an error because it was unable to return all outstanding responses (we observe this with a `MessageReceiveTimeout`)

Importantly, if `get_all_responses` does not raise an error, the response to this request _will_ be in the return value somewhere (and 99.97% of the time it our case, it's the first and only response in the list). Any other responses returned are hopelessly lost anyway. Their request IDs were discarded when the they previously erred out, so it is impossible to match those other responses to their previous requests. Discarding them is the only proper behavior, and there is no need to try to save them (as this pull request does).

pysoa/client/client.py
sethbrite: It'll only "never get consumed" by clients calling `get_response_for_request`, which is to be expected.

pysoa/client/client.py
sethbrite: We're building a list of unexpected responses we see (as opposed to the expected response that we want to retrieve). At the very end of this function, we set the unexpected responses we saw to the list of responses we should return first via `_get_responses` (vs. getting new ones from the transport).

sethbrite:
@nickwilliams-eventbrite but what if there is already a response on the queue? Say, a response from a previously timed out request (i.e. the initial cause of the error this diff solves). In that case, there's no guaranteeing that the responses received will contain the expected response for the request. It's actually highly unlikely, since the request was _just_ made. For Redis transport, this would mean that the server would have to have pushed a `JobResponse` to the appropriate list before the client does a `BLPOP` mere microseconds later.

nickwilliams-eventbrite:
>@nickwilliams-eventbrite but what if there is already a response on the queue? Say, a response from a previously timed out request (i.e. the initial cause of the error this diff solves). In that case, there's no guaranteeing that the responses received will contain the expected response for the request. It's actually highly unlikely, since the request was just made. For Redis transport, this would mean that the server would have to have pushed a JobResponse to the appropriate list before the client does a BLPOP mere microseconds later.

@sethbrite, it's not highly unlikely. It's _guaranteed_. `get_all_responses` will _block_ until the response for the request you just sent is received (or raise an error if it times out). _Even if other responses are received first, your response will be somewhere in the received responses (unless it times out)._

sethbrite:
@nickwilliams-eventbrite what about `get_all_responses` guarantees that you get the response for the request you just made? Maybe I'm missing something, but it looks to me like it gets _any_ responses from the transport.

https://github.com/eventbrite/pysoa/blob/master/pysoa/client/client.py#L162-L170

nickwilliams-eventbrite:
>@nickwilliams-eventbrite what about get_all_responses guarantees that you get the response for the request you just made? Maybe I'm missing something, but it looks to me like it gets any responses from the transport.
>
>https://github.com/eventbrite/pysoa/blob/master/pysoa/client/client.py#L162-L170

`get_all_responses` doesn’t just return one response or just all of the responses that are already available. It keeps returning responses until the number of responses it has returned is equal to the number of requests it has sent. If it has returned fewer responses than requests it sent, it will block until another response is available or it times out. It is not possible for the method to complete without error if there are still outstanding responses.

To dive into it a bit more technically, the _client_ transport method `receive_response_message` is contractually obligated to do exactly one of two things, no more or less:

- Return `None, None, None` if it has already received as many responses as it has sent (so it's not possible for any more responses to come back)
- Otherwise, block until it receives the next response, or time out if it waits too long

`get_all_responses` is a generator that loops indefinitely, yielding responses returned by `receive_response_message`, until `receive_response_message` returns `None, None, None`, at which point `get_all_responses` raises `StopIteration` (it returns/completes). Thus, it is not possible for `get_all_responses` to complete without error if there are still outstanding responses.

pysoa/test/stub_service.py
```diff
@@ -370,6 +371,7 @@ def wrapped_send_request(client, service_name, actions, *args, **kwargs):
                     request_id,
                     continue_on_error,
                 )
+                self._stub_action_unwrapped_call_id_map[request_id] = unwrapped_request_id
```
nickwilliams-eventbrite: You don't need these two lines with `_stub_action_unwrapped_call_id_map` anymore.

pysoa/client/client.py
```diff
@@ -34,6 +34,7 @@ class ServiceHandler(object):
     _transport_cache = {}
 
     def __init__(self, service_name, settings):
+        self.service_name = service_name
```
nickwilliams-eventbrite: You don't need this set anymore.

pysoa/client/client.py
```diff
@@ -340,12 +342,27 @@ def call_actions(self, service_name, actions, expansions=None, raise_action_erro
         if timeout:
             kwargs['message_expiry_in_seconds'] = timeout
 
-        request_id = self.send_request(service_name, actions, **kwargs)
-        # Dump everything from the generator. There should only be one response.
+        expected_request_id = self.send_request(service_name, actions, **kwargs)
```
nickwilliams-eventbrite: So, you changed this to `expected_request_id`, and then changed below from `response_id` to `request_id`. I'm not dead set against that, but I'm not sure it makes sense. The value returned from `send_request` _is_ the request ID, and the value returned from `get_all_responses` _is_ the response ID (and the two are supposed to match). If it were me, I'd leave it `request_id` and `response_id`, but I won't hold it up.

PR #87, opened 2018-04-21
[MAJOR] Add extensive test plan system with customized test plan syntax

nickwilliams-eventbrite:
The PySOA Test Plan system allows for writing streamlined action-testing test plans, fixtures, and test cases to make testing your services easier. It comes with a base class for all test plans, a parser for the syntax, a PyTest plugin, and an extensive set of sytax directives meeting most needs. Our internal SOA system will add more directives for handling our proprietary auth tokens, and users are encouraged to do the same for their services.


adamsussman:
Awesome job porting a large code base.  Some questions:

1) The  doc generator will always be complete unless other modules that provide grammar plugins are also installed and present.  How is this intended to be used?

2) Speaking of the other modules, have any of those been ported/tested under the new structure (The django model and gargoyle ones are probably critical and complicated).

3) Are there any aids to the conversion of old soa test plans to pysoa test plans?

pysoa/test/plan/parser.py
```diff
@@ -0,0 +1,289 @@
...
+        self._ingest_directives()
+
+        if not self.test_cases:
+            raise FixtureLoadError('Failed to find any test cases in file: {}'.format(self._fixture_file_name))
+
+        return self.test_cases
+
+    def _ingest_directives(self):
+        try:
+            with codecs.open(self._fixture_file_name, mode='rb', encoding='utf-8') as file_input:
```
sethbrite: Why use `codecs.open`? Should we use `io.open` instead?

nickwilliams-eventbrite: A while back, I did a bunch of research about what is the best way to open a file with an encoding that would work exactly the same in both Python 2 and Python 3, and the majority opinion leaned heavily toward `codecs.open` (since straight `open` doesn't support an encoding until Python 3.x). As such, that's the way I have done it in all of our Python 2- and 3-compatible code.

pysoa/test/plugins/pytest.py
```diff
@@ -0,0 +1,535 @@
...
+            line_number=self.lineno + 1,
+            test=self.name,
+            source=self._test_source,
+        )
+
+    def __repr__(self):
+        return '<TracebackEntry {}:{}>'.format(self.path, self.lineno + 1)
+
+
+def pytest_collection_modifyitems(config, items):
```
flub: This sounds like it should be possible to do with normal marking instead.  Maybe the collector or something else can use applymarker() to mark the tests and then more usual mechanisms can be used to select these tests instead of having to implement all this yourself.

PR #92, opened 2018-05-03
[PATCH] Ensure an error response is sent if response too large

nickwilliams-eventbrite:
If a serialized response message is too large to send, resulting in a `MessageTooLarge` error, the error was trickling back all the way up to the server process loop, causing the server to shut down and the client to time out waiting for a response. This change ensures that, in such a case, an error response is sent to the client informing it that the response was too large to send, and then the server continues on with its business as usual instead of shutting down.

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -12,12 +12,14 @@ import signal
 import attr
 from pysoa.client import Client
 from pysoa.common.constants import (
+    ERROR_CODE_RESPONSE_TOO_LARGE,
     ERROR_CODE_SERVER_ERROR,
     ERROR_CODE_UNKNOWN,
 )
 from pysoa.common.transport.exceptions import (
     MessageReceiveError,
     MessageReceiveTimeout,
+    MessageTooLarge,
 )
 from pysoa.common.types import (
     ActionResponse,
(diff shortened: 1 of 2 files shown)


pysoa/server/server.py
```diff
@@ -108,29 +110,50 @@ def handle_next_request(self):
...
-                    # When we originally logged the request, it may have been hidden because the effective logging level
-                    # threshold was greater than the level at which we logged the request. So re-log the request at the
-                    # error level, if set higher.
-                    self.job_logger.log(self.request_log_error_level, 'Job request: %s', request_for_logging)
-                self.job_logger.log(self.request_log_error_level, 'Job response: %s', response_for_logging)
-            else:
-                self.job_logger.log(self.request_log_success_level, 'Job response: %s', response_for_logging)
+            # Send the response message
+            try:
+                self.transport.send_response_message(request_id, meta, response_message)
```
yanguo-eb: How is the response size limit configed? Could we check the size of `response_message` here instead of `try`?

nickwilliams-eventbrite: We cannot. The response size limit is driven by the transport. (The local transport, for example, doesn't have a limit, but the Redis transport has a 100KB limit.) We have to just wait for the transport to complain.

PR #97, opened 2018-05-09
[MINOR] Add a SwitchedAction class to facilitate switch usage

nickwilliams-eventbrite:
Add a `SwitchedAction` class, which facilitates using different actions for a single given action name depending on which switch(es), if any, were sent with the request context.

### DIFF
--- a/pysoa/server/action/switched.py
+++ b/pysoa/server/action/switched.py
@@ -1,0 +1,145 @@
+from __future__ import absolute_import, unicode_literals
+
+import abc
+
+import six
+
+from pysoa.server.internal.types import is_switch
+
+
+__all__ = (
+    'SwitchedAction',
+)
+
+
+def _len(item):
+    # Safe length that won't raise an error on values that don't support length
+    return getattr(item, '__len__', lambda *_: -1)()
+
+
+class _DefaultAction(object):
+    def __int__(self):
+        d = id(self)
+        return d if d < 0 else -d
+
+    def __eq__(self, other):
+        return getattr(other, '__class__', None) == _DefaultAction
+
+
+class _SwitchedActionMetaClass(abc.ABCMeta):
+    def __new__(mcs, name, bases, body):
+        """
+        Validate the switch_to_action_map when the class is created, instead of doing it every time the class
+        is instantiated. This identifies problems earlier (on import) and improves performance by not performing this
+        validation every time the action is called.
+        """
+        cls = super(_SwitchedActionMetaClass, mcs).__new__(mcs, name, bases, body)
+
+        # noinspection PyUnresolvedReferences
+        if bases[0] is not object and (
+            not cls.switch_to_action_map or
+            not hasattr(cls.switch_to_action_map, '__iter__') or
+            _len(cls.switch_to_action_map) < 2 or
+            any(
+                True for i in cls.switch_to_action_map
+                if not hasattr(i, '__getitem__') or _len(i) != 2 or not is_switch(i[0]) or not callable(i[1])
+            )
+        ):
+            raise ValueError(
+                'Class attribute switch_to_action_map must be an iterable of at least two indexable items, each '
+                'with exactly two indexes, where the first element is a switch and the second element is an action '
+                '(callable).'
+            )
+
+        return cls
+
+
+@six.add_metaclass(_SwitchedActionMetaClass)
+class SwitchedAction(object):
+    """
+    A specialized action that defers to other, concrete actions based on request switches. Subclasses must not
+    override any methods and must override `switch_to_action_map`. `switch_to_action_map` should be some iterable
+    object that provides `__len__` (such as a tuple [recommended] or list). Its items must be indexable objects that
+    provide `__len__` (such as a tuple [recommended] or list) and have exactly two elements.
+
+    For each item in `switch_to_action_map`, the first element must be a switch that provides `__int__` (such as an
+    actual integer) or a switch that provides an attribute `value` which, itself, provides `__int__` (or is an int).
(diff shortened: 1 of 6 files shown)


pysoa/server/action/switched.py
```diff
@@ -0,0 +1,122 @@
...
+    For each item in `switch_to_action_map`, the first element must be a switch that provides `__int__` (such as an
+    actual integer) or a switch that provides an attribute `value` which, itself, provides `__int__` (or is an int).
+    The second element must be an action, such as an action class (e.g. one that extends `Action`) or any callable
+    that accepts a server settings object and returns a new callable that, itself, accepts an `ActionRequest` object
+    and returns an `ActionResponse` object or raises an `ActionError`.
+
+    `switch_to_action_map` must have at least two items in it. `SwitchedAction` will iterate over that list, checking
+    the first element (switch) to see if it is enabled in the request. If it is, the second item (the action) will be
+    deferred to. If it finds no items whose switches are enabled, it will use the very last action in
+    `switch_to_action_map`. As such, you can treat the last item as a default, and its switch could simply be `0`.
```
andrewgodwin: I would rather we had an explicit default rather than it being the last one - say that "value X means default", and have it be last be a convention rather than required. I might suggest `None` rather than `0` as well.

pysoa/server/internal/types.py
```diff
@@ -1,76 +1,80 @@
-import six
+from __future__ import absolute_import, unicode_literals
+
+
+def get_switch(item):
+    if hasattr(item, '__int__'):
+        return item.__int__()
+    if hasattr(getattr(item, 'value', None), '__int__'):
+        return item.value.__int__()
+    return None
```
andrewgodwin: I'd raise an error here rather than `None` - the `TypeError` from below, in fact.

pysoa/server/action/switched.py
nickwilliams-eventbrite: Updated, as discussed here and on Slack.

pysoa/server/internal/types.py
nickwilliams-eventbrite: Changed.

pysoa/server/action/switched.py
```diff
@@ -0,0 +1,126 @@
...
+
+        :param settings: The server settings object
+        :type settings: dict
+        """
+        if self.__class__ is SwitchedAction:
+            raise TypeError('Cannot instantiate abstract SwitchedAction')
+
+        self.settings = settings
+
+    def get_raw_action(self, action_request):
```
vartec: I'd call that `get_action_class` (even though in theory it can be something that is not a Python class, it's still used as such)

pysoa/server/action/switched.py
```diff
@@ -0,0 +1,126 @@
...
+            ...
+
+        class UserTransitionAction(SwitchedAction):
+            switch_to_action_map = (
+                (USER_VERSION_2_ENABLED, UserActionV2),
+                (SwitchedAction.DEFAULT_ACTION, UserActionV1),
+            )
+    """
+
+    DEFAULT_ACTION = -1827739192
```
vartec: any reason why not just `-1`, given that switches are positive integers?

pysoa/server/action/switched.py
```diff
@@ -0,0 +1,126 @@
...
+
+        :param action_request: The request object
+        :type action_request: EnrichedActionRequest
+
+        :return: The action
+        :rtype: callable
+        """
+        action = None
+
+        for item in self.switch_to_action_map:
```
vartec: `for switch, action in self.switch_to_action_map` instead of using `item[x]` would probably be more readable

pysoa/server/action/switched.py
nickwilliams-eventbrite: I have renamed it `get_uninitialized_action`.

pysoa/server/action/switched.py
nickwilliams-eventbrite: _Our_ switches are positive integers. There's no hard requirement in SOA that _all_ switches be positive. The only requirement is "integer."

pysoa/server/action/switched.py
nickwilliams-eventbrite: I had it this way originally. I changed it for a reason: I only wanted to require that `__getitem__` was provided. As it turns out, `__getitem__` is enough for tuple-style unpacking to work. So I've changed it back to your suggestion.

pysoa/server/action/switched.py
vartec: OK, so the idea is that it's just a random number, that by chance will not be used?
Then I vote for `-802549514462` ;-)

andrewgodwin: Is there a reason this has to be an `int`, rather than something obviously unique like a brand new class (what `NOT_PROVIDED` is powered by in Django) or `None`?

vartec: It can't be as trivial as this unfortunately 
```In [1]: class Foo:
   ...:     def __int__(self):
   ...:         return None
   ...:

In [2]: foo = Foo()

In [3]: int(foo)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-3bf93630f200> in <module>()
----> 1 int(foo)

TypeError: __int__ returned non-int (type NoneType)```

Comment (vartec) [reply pysoa/server/action/switched.py]: This will work though:

```
class Foo(int):
    def __eq__(self, other):
        return False
```

nickwilliams-eventbrite: Alright. Attempt 3. How bow now?

PR #99, opened 2018-05-11
[MINOR] Remove the transport cache as it is no longer needed

nickwilliams-eventbrite:
The transport cache (controlled by the `transport_cache_time_in_seconds` setting), was put in place before we got our Redis connection handling under control, when creating new transports took multiples of milliseconds. Now that we properly handle Redis connections with a thread-safe, global connection cache, creating new transports take nanoseconds, the same as retrieving transports from the cache. Additionaly, the cache is not thread-safe and introduces problems in multi-threaded environments.

As such, we are removing the transport cache from PySOA. This is a backwards-compatible change, because we are simply deprecating the `transport_cache_time_in_seconds` setting, and leaving it in place for some time. There will be a future, breaking change that removes this setting altogether. Currently, its value is completely ignored.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -621,7 +621,7 @@ Settings Schema Definition
     Optional keys: ``kwargs``
 
 
-- ``transport_cache_time_in_seconds`` - ``integer``: If enabled, uses a per-service transport cache that is keyed off the service name and transport settings, persists across all clients in memory, and expires after this number of seconds. By default, a new transport is created for every new client. Note that this is safe in a multi-processing environment, but not in a multi-threaded environment. You should disable this in a multi-threaded environment. (additional information: ``{u'gte': 0}``)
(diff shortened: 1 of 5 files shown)


docs/reference.rst
```diff
@@ -621,7 +621,7 @@ Settings Schema Definition
     Optional keys: ``kwargs``
 
 
-- ``transport_cache_time_in_seconds`` - ``integer``: If enabled, uses a per-service transport cache that is keyed off the service name and transport settings, persists across all clients in memory, and expires after this number of seconds. By default, a new transport is created for every new client. Note that this is safe in a multi-processing environment, but not in a multi-threaded environment. You should disable this in a multi-threaded environment. (additional information: ``{u'gte': 0}``)
+- ``transport_cache_time_in_seconds`` - ``anything``: This field is deprecated. The transport cache is no longer supported. This settings field will remain in place for a month or so to give a safe period for people to remove it from settings, but its value will always be ignored.
```
vartec: Whoever is gonna read that in the future, won't know `a month or so` since when (w/o having to look into git lot)

nickwilliams-eventbrite: Good point! I have reworded this with an exact date.

PR #101, opened 2018-05-31
[MINOR] Add support for decimal.Decimal in MessagePack serializer

nickwilliams-eventbrite:


pysoa/common/serializer/msgpack_serializer.py
```diff
@@ -16,32 +19,53 @@ class MsgpackSerializer(BaseSerializer):
...
+    5. Currint amount and currency
+
+       Encoded as a 3-byte ASCII string of the uppercased currency code concatenated with a 8-byte big-endian signed
+       integer of the minor value.
     """
 
     mime_type = u'application/msgpack'
 
-    EXT_DATETIME = 1
+    EXT_CURRINT = 2
```
vartec: I know PEP8 suggest sorting these by name, but I feel it _would_ make it more readable to sort this block by value

pysoa/common/serializer/msgpack_serializer.py
```diff
@@ -85,6 +109,14 @@ def default(self, obj):
                 self.EXT_TIME,
                 self.STRUCT_TIME.pack(obj.hour, obj.minute, obj.second, obj.microsecond),
             )
+        elif isinstance(obj, decimal.Decimal):
+            obj_str = six.text_type(obj)[:65535].encode('utf-8')
```
vartec: shouldn't this raise a `TypeError` if length > 65535?

nickwilliams-eventbrite: Remember, you're talking about a number with more than 65,535 _digits_. Not a number greater than 65,535. The number would either have to be greater than `10^65534` or have a post-decimal precision of `10^-65534`, or some combination of the two. In other words, this will never happen. However...

The built-in string pack types truncate if the string is longer than the length supported by the leading length byte(s), so this mimics that behavior.

pysoa/common/serializer/msgpack_serializer.py
nickwilliams-eventbrite: Yeah, I went back and forth on this. Settled on alphabetization, because the `STRUCT_**` constants below these constants _only_ make sense alphabetical, and it makes sense for these related constants to have the same order as the `STRUCT_**` constants to which they relate.

pysoa/common/serializer/msgpack_serializer.py
vartec: I agree it's very unlikely to happen with real data. I don't like silent failures. If it'd be protected by an exception, it would be unlikely to be raised with real data. Not gonna block on that though.

PR #104, opened 2018-06-14
[MAJOR] Fix bug causing server to shut down on unserializable responses

nickwilliams-eventbrite:
If an action returns a response which the transport cannot serialize, the server was failing hard and shutting down, and the client never received an error response (instead, it timed out). This change catches the error and handles it cleanly.


nickwilliams-eventbrite:
>do we need something similar on the client side?

Nope. Client side will handle it like any other job error, as it should. The point is that the client side will _get_ an error now, instead of timing out.

pysoa/server/server.py
```diff
@@ -255,11 +266,11 @@ def process_job(self, job_request):
             # Send an error response if no middleware caught this.
             # Formatting the error might itself error, so try to catch that
             self.metrics.counter('server.error.unhandled_error').increment()
-            return self.handle_error(e)
+            return self.handle_exception(e)
 
         return job_response
 
-    def handle_error(self, error, variables=None):
+    def handle_exception(self, error, variables=None):
```
yanguo-eb: Is it better to keep this method name unchanged? Also name the new method some other name other than `*error` or `*exception`, as those two words interleave and confusing.

nickwilliams-eventbrite: I renamed that method specifically because it was ambiguous, but I just ended up with two ambiguous method names. I'll improve further.

nickwilliams-eventbrite: How does it look now, Yan?

PR #108, opened 2018-07-13
adding support for errors due insufficient permissions

vartec:
This additional field would let Actions indicate that a permission is needed, but the authenticated user does not have it. 

This is free form, and does not depend on any specific implementation of permissions system.


nickwilliams-eventbrite:
Please squash-and-merge. Don't merge all three separate commits into `master`.

PR #109, opened 2018-07-13
[MINOR] Add option to suppress responses for send-and-forget

nickwilliams-eventbrite:
We've started using a send-and-forget concept, where we send a request but neither care about nor wait on a response. In these cases, the response the server sends needlessly sits in the Redis queue for 60 seconds until it expires. Using a `suppress_response` flag, you can now implement send-and-forget and instruct the server to not send a response, in order to conserve resources.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -231,7 +231,7 @@ class Client(object):
             errors_string = '\n'.join(['{a.action}: {a.errors}'.format(a=a) for a in self.actions])
             return 'Error calling action(s):\n{}'.format(errors_string)
 
-    # Synchronous request methods
+    # Blocking methods that send a request and wait until a response is available
 
     def call_action(
         self,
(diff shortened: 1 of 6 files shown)


pysoa/client/client.py
```diff
@@ -570,6 +570,90 @@ def call_jobs_parallel(
 
         return responses
 
+    # Methods used to send a request in a non-blocking manner and then later block for a response as a separate step
+
+    def send_request(
```
nickwilliams-eventbrite: This method is not new. It was moved from the bottom of the class to here, above all the private methods.

pysoa/client/client.py
```diff
@@ -570,6 +570,90 @@ def call_jobs_parallel(
...
+        )
+        context = self._make_context_header(
+            switches=switches,
+            correlation_id=correlation_id,
+            context_extra=context,
+        )
+        job_request = JobRequest(actions=actions, control=control, context=context or {})
+        return handler.send_request(job_request, message_expiry_in_seconds)
+
+    def get_all_responses(self, service_name, receive_timeout_in_seconds=None):
```
nickwilliams-eventbrite: This method is not new. It was moved from the bottom of the class to here, above all the private methods.

PR #110, opened 2018-07-13
[MAJOR] Make the maximum Redis transport message size configurable

nickwilliams-eventbrite:
The maximum message size for the Redis Gateway transport was not set to a reasonable amount. This change makes it a configurable setting instead of a hard-coded value, and sets more reasonable defaults for clients and servers, separately. It also does some log refactoring to improve logging around message size. In particular:

- BREAKING: `pysoa.server.logging` has moved to `pysoa.common.logging` so that it can be used by the transport layer
- `MessageTooLarge` errors now include the message size
- The message size is logged when a messaged fails to send due to being oversized
- You can now configure a lower threshold at which to log, but still send, less-oversized messages


pysoa/common/transport/redis_gateway/core.py
```diff
@@ -194,7 +199,7 @@ def send_message(self, queue_name, request_id, meta, body, message_expiry_in_sec
         with self._get_timer('send.serialize'):
             serialized_message = self.serializer.dict_to_blob(message)
 
-        if len(serialized_message) > self.MAXIMUM_MESSAGE_BYTES:
+        if len(serialized_message) > self.maximum_message_size_in_bytes:
```
vartec: could we additionally do
```
elif len(serialized_message) > (self.maximum_message_size_in_bytes * .9):
    log.warn('large message') # or something like that
```

(this could be future improvement, by no means blocking this change)

PR #113, opened 2018-07-25
[MAJOR] Add support for non-blocking client futures

nickwilliams-eventbrite:
Now you can use `_future` versions of the `Client`'s `call_` methods to send requests and then perform other activities before obtaining the response using the returned future.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -78,6 +78,25 @@ Returns
 Raises
   ``ConnectionError``, ``InvalidField``, ``MessageSendError``, ``MessageSendTimeout``, ``MessageTooLarge``, ``MessageReceiveError``, ``MessageReceiveTimeout``, ``InvalidMessage``, ``JobError``, ``CallActionError``
 
+.. _pysoa.client.client.Client.call_action_future:
+
+``method call_action_future(service_name, action, body=None, **kwargs)``
+************************************************************************
+
+This method is identical in signature and behavior to ``call_action``, except that it sends the request and
+then immediately returns a ``FutureResponse`` instead of blocking waiting on a response and returning
+an ``ActionResponse``. Just call ``result(timeout=None)`` on the future response to block for an available
+response. Some of the possible exceptions may be raised when this method is called; others may be raised when
+the future is used.
+
+Parameters
+  - ``service_name``
+  - ``action``
+  - ``body``
+
+Returns
+  ``Client.FutureResponse`` - A future from which the action response can later be retrieved
+
 .. _pysoa.client.client.Client.call_actions:
 
 ``method call_actions(service_name, actions, expansions=None, raise_job_errors=True, raise_action_errors=True, timeout=None, **kwargs)``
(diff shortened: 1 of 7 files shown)


andrewgodwin:
I'm curious why you chose to make your own classes than just returning a classic Python `Future`? The "request an attribute and the whole process will wait for the response" part of this seems a little like soft-failure, I'd sort of rather we made people interact with Futures the Python way and then touch the resultant object once it was fully materialised.

vartec:
@andrewgodwin but what do you mean by `Python Future` in case of Python 2.7?

andrewgodwin:
My bad, I confused early python 3 with 2. I would still be curious if we could replicate 3's concurrent.future more though.

nickwilliams-eventbrite:
>I'm curious why you chose to make your own classes than just returning a classic Python `Future`? ... I'd sort of rather we made people interact with Futures the Python way and then touch the resultant object once it was fully materialised.

>@andrewgodwin but what do you mean by `Python Future` in case of Python 2.7?

>My bad, I confused early python 3 with 2. I would still be curious if we could replicate 3's concurrent.future more though.

Well, there is a `concurrent.future` backport for Python 2 (just like `ipaddress` and `enum`). However, the problem with Python's concurrent `Future` is that it (and its API) is specifically designed for some thing that gets executed by an `Executor` (in another thread) and then you can get its results after it's done. It's only barely different from `threading.Thread`, and it's wildly different from what we're trying to do here: Send a request to a service for processing and return immediately, returning an easy interface (a future, in the generic sense) to get the response for that request at a later time. Involving threads and executors is unnecessary (this is handled by the service existing remotely). Additionally, it's a level of complexity that distracts from the desired simplicity of this approach. 

I did, initially, try to make this work with something built in to Python, like `Future`, but there just wasn't anything that was close enough to what we needed that I felt happy with using.

>The "request an attribute and the whole process will wait for the response" part of this seems a little like soft-failure

Understandable. To be clear, though, it's clearly documented that way (although, I could improve upon this documentation even further, if you desired).

andrewgodwin:
Well, the `Future` abstraction in Python 3 works for both the old-style executors in `concurrent` and also for the async stuff in `asyncio`, so it's not like it's specifically designed for thread-based execution - it's designed around the standard Promise/Future pattern a lot of modern languages are going for (and that Python has had since Twisted).

Cruicially, the key part of that is that a Future only has a few attributes - the ones I care about are `.result()`, which blocks and returns the result (or raises an exception if needed), and `.exception()`, which blocks and tells you if a call failed or not.

Having explicit accessors like this means you can tell you are reading future-based code when you look at it, rather than the lazy-style accessors you have implemented, where you have to know that the variable you're looking at is something that might block on attribute access where you might not be expecting it. That's why I'd generally prefer something in the same model, where the returned future just has something that either returns a response object or an exception.

nickwilliams-eventbrite:
@vartec, @miloops, @sethbrite, what do y'all think? Here are the competing proposals:

Option 1: The way futures are currently implemented in this PR

Option 2: Something more like this, without API-specific futures:

```
    class FutureResponse(...):
        ...
        def result(self):
            # logic to get and return response or return cached response or raise error
```

sethbrite:
@nickwilliams-eventbrite I think if we're going to call something a Future in Python, I think it should have the common Python Future API.

nickwilliams-eventbrite:
Okay, @andrewgodwin, @sethbrite, and @vartec: I have revised this to more closely match the `concurrent.Future` API. It doesn't have `send`, `cancel`, `cancelled`, `add_done_callback`, or any of the `set_*` methods, because those have no relevance to what we're doing here.

PR #117, opened 2018-08-07
Pass position in job to action for multi-action jobs

adamsussman:
When there are multiple actions in a request, it is sometimes useful for the action to know its position in the job.


nickwilliams-eventbrite:
When is it sometimes useful? What is the use case for this?

adamsussman:
I refer you to the variety of old soa services that work in multiple action per job requests.  Most of the ones I know about are financial in nature.  In any case, this is needed for old soa emulation by pysoa, which will be along in a review shortly.

PR #121, opened 2018-08-14
[MINOR] Improve logging configuration to not conflict with Django

nickwilliams-eventbrite:
If Django `settings.LOGGING` is present, Django will call `logging.config.dictConfig`. This can conflict with PySOA's call to `logging.config.dictConfig` with `settings.SOA_SERVER_SETTINGS['logging']`. This change ensures that the two logging configs are identical so that we'll have consistent behavior.

### DIFF
--- a/pysoa/server/standalone.py
+++ b/pysoa/server/standalone.py
@@ -3,6 +3,8 @@ from __future__ import (
     unicode_literals,
 )
 
+import importlib
+import logging
 import sys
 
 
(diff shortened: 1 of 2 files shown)


pysoa/server/standalone.py
```diff
@@ -164,6 +165,20 @@ def django_main(server_getter):
     if args.settings:
         os.environ['DJANGO_SETTINGS_MODULE'] = args.settings
 
+    try:
+        # We have to import it manually, because we need to manipulate the settings before setup() is called, but we
+        # can't import django.conf.settings until after setup() is called.
+        django_settings = importlib.import_module(os.environ['DJANGO_SETTINGS_MODULE'])
+        django_settings.LOGGING = django_settings.SOA_SERVER_SETTINGS['logging']
```
vartec: so this is preserved even if module is imported again by other code?

pysoa/server/standalone.py
```diff
@@ -164,6 +165,20 @@ def django_main(server_getter):
...
+    except ImportError:
+        raise ValueError('Cannot import Django settings module `{}`.'.format(os.environ['DJANGO_SETTINGS_MODULE']))
```
yanguo-eb: When `args.settings` is None, may this line throw another exception because `os.environ['DJANGO_SETTINGS_MODULE']` is undefined?

pysoa/server/standalone.py
nickwilliams-eventbrite: Yes.

pysoa/server/standalone.py
nickwilliams-eventbrite: No. `os.environ['DJANGO_SETTINGS_MODULE']` can never be undefined when we reach this point in code. This environment variable is either already defined _or_ `args.settings` is made a required parameter and the code errors before it gets here if the user doesn't provide that argument.

nickwilliams-eventbrite:
>What if LOGGING is already defined? Should we raise an exception advising to put those in the right place? So we don't silently override them.

Good point, @miloops. So, if `LOGGING` is defined and differs from `SOA_SERVER_SETTINGS['logging']`, you would prefer it to raise an exception and exit, rather than override `LOGGING`? This is solution 3B discussed on Slack yesterday. It differs from solution 3C, which you proposed and everyone seemed to agree with. Just wanted to make sure we're on the same page.

>Nick: 3B: Implement a check in PySOA that errors out if `LOGGING` differs from `SOA_SERVER_SETTINGS`, in order to enforce this rule.
>
>Emilio: Could PySOA set `LOGGING = SOA_SERVER_SETTINGS['logging']` ? And just people should just set `SOA_SERVER_SETTINGS['logging']` and warn if `LOGGING` is defined?
>
>Nick: That’s a possible 3C, sure.
>
>Emilio: Then my vote if for 3C 🙂

miloops:
>Good point, @miloops. So, if LOGGING is defined and differs from SOA_SERVER_SETTINGS['logging'], you would prefer it to raise an exception and exit, rather than override LOGGING? This is solution 3B discussed on Slack yesterday. It differs from solution 3C, which you proposed and everyone seemed to agree with. Just wanted to make sure we're on the same page.

That's fair, but we are not warning either in this diff, or I'm not seeing it. Should we add that?

nickwilliams-eventbrite:
>That's fair, but we are not warning either in this diff, or I'm not seeing it. Should we add that?

We can add a warning. The concern is that, at this point, logging isn't set up yet, because settings haven't been loaded yet. I might have to do some clever things. I'll think on it a bit.

nickwilliams-eventbrite:
@miloops, how does it look now?

PR #122, opened 2018-08-15
Add tools to support pytesty testing in pysoa services

ebwill:
Introduce a few fixtures and contextmanagers to enable testing using
pytest's full suite of features. See https://docs.google.com/document/d/1AHeTmRZf-x2cpQ67Rw4oY8gFb88FgSMXVG3MTTe2tmk/edit?usp=sharing
for more info.


ebwill:
@vartec yes

pysoa/test/pytest_fixtures.py
```diff
@@ -0,0 +1,71 @@
...
+        from django.conf import settings
+        settings = settings.SOA_SERVER_SETTINGS
+    else:
+        settings_module = os.environ.get('PYSOA_SETTINGS_MODULE', None)
+        assert settings_module, 'PYSOA_SETTINGS_MODULE environment variable must be set to run tests.'
+        try:
+            thing = importlib.import_module(settings_module)
+            settings = thing.SOA_SERVER_SETTINGS
+        except (ImportError, AttributeError) as e:
+            raise AssertionError('Could not access {}.SOA_SERVER_SETTINGS: {}'.format(settings_module, e))
```
nickwilliams-eventbrite: So, in the `Server` class, we [permit this non-Django settings module to have _either_ `SOA_SERVER_SETTINGS` _or_ `settings`](https://github.com/eventbrite/pysoa/blob/4f86edea4f535d137ea575d38abdd664381a2774/pysoa/server/server.py#L643-L653). Perhaps we should do the same here.

pysoa/test/pytest_fixtures.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.fixture(scope='session')
+def call_action(server_class, service_client):
+    """
+    Fixture to call a service action. Defaults to calling the localized service.
+    """
+    def do_call_action(action, body=None, service_name=None, **kwargs):
+        return service_client.call_action(service_name or server_class.service_name, action, body=body, **kwargs)
+
+    return do_call_action
```
nickwilliams-eventbrite: As Diego mentioned in the Phab diff for EB SOA, this is going to be problematic in PyTest 4. Additionally, I'm uncertain about the fixture name as it is, and definitely against the fixture name if necessary changes are made to be compatible with PyTest 4. This will require some changes.

I have a more foundational concern here, though. Why do we need a fixture for `call_action` at all? This is not really following the paradigm of PyTest fixtures. The `service_client` fixture should be sufficient, and people can just call `call_action` (or whatever) on that fixture (which is the way fixtures are meant to be used). I'm just not convinced we need this fixture at all.

setup.py
```diff
@@ -67,6 +67,7 @@ def readme():
     entry_points={
         'pytest11': [
             'pysoa_test_plan=pysoa.test.plugins.pytest',
+            'pysoa_test_fixtures=pysoa.test.pytest_fixtures',
```
nickwilliams-eventbrite: If `pytest_fixtures` is handled like a PyTest plugin, then it should go in `pysoa.test.plugins`. We might need to move `pysoa/test/plugins/pytest.py` to `pysoa/test/plugins/pytest/plans.py` and create `pysoa/test/plugins/pytest/fixtures.py`.

pysoa/test/pytest_fixtures.py
ebwill: Diego was mistaken in the other diff (I replied there), this pattern is supported by Pytest 4. The reason I added this fixture is to enable the shortcut we use in our tests of dropping the service name, both for convenience and ease of migration between unittest-style tests and pytest-style. It's the pytest equivalent of the `call_action` method on `ServerTestCase`.

ebwill: My favorite part about it is enabling the addition of the `call_action_anon` fixture in ebsoa, as it makes it easy for devs to do the right thing (ie not generate users just to call actions, but to stub permissions, and to use an auth_token when calling actions anonymously)

setup.py
ebwill: cool, will move there

pysoa/test/server.py
```diff
@@ -7,9 +7,8 @@
 import os
 import unittest
 
-import six
-
 from pysoa.client import Client
+from pysoa.test.assertions import raises_call_action_error, raises_field_errors, raises_error_codes
```
mauricioh-eb: nit, line too long?

pysoa/test/plugins/fixtures.py
```diff
@@ -0,0 +1,93 @@
...
+    unicode_literals,
+)
+
+import importlib
+import os
+
+import pytest
+
+
+@pytest.fixture(scope='session')
```
nickwilliams-eventbrite: I had a thunk. Why session scope? This makes it difficult to override just one fixture on a per-module basis, because then you also have to override all the other fixtures that depend on it. I would argue `scope='module'` or even `scope='function'` is more appropriate.

pysoa/test/plugins/fixtures.py
```diff
@@ -0,0 +1,93 @@
+from __future__ import (
```
nickwilliams-eventbrite: `plugins/pytest/fixtures.py`, please. Not `plugins/fixtures.py`. Also, below, `plugins/pytest/plans.py`, please. Not `plugins/plans.py`.

pysoa/test/server.py
nickwilliams-eventbrite: This must be a multi-line import. Please run `isort` and resolve all issues (except one in an unrelated file where it tries to move the location of a comment within imports, which is an `isort` bug).

pysoa/test/plugins/fixtures.py
ebwill: so many directories

pysoa/test/plugins/fixtures.py
ebwill: `scope` is kind of a misnomer, it just defines the lifetime of the fixture. `session` means that the fixture will be evaluated at most once per test run, but if the fixture itself lives in a test module, it will only be available to tests in that module, and will only override equivalently named fixtures in the scope of that module.

PR #125, opened 2018-08-31
[MINOR] Make the polymorphic client and server settings extensible

nickwilliams-eventbrite:
Before this change, if a consumer of PySOA (such as a service or another library) provides a new transport with a validated settings schema, the only way to use the transport and ensure the settings for it are validated is to also extend/override `PolymorphicServerSettings` and `PolymorphicClientSettings` and extend the `Client` to use the new transport settings class. This change makes the polymorphic client and server settings extensible by adding a `settings_schema` class attribute to a transport class.

### DIFF
--- a/pysoa/common/schemas.py
+++ b/pysoa/common/schemas.py
@@ -5,6 +5,8 @@ from __future__ import (
 
 from conformity import fields
 
+from pysoa.utils import resolve_python_path
+
 
 class BasicClassSchema(fields.Dictionary):
     contents = {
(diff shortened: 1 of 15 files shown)


pysoa/common/schemas.py
```diff
@@ -34,3 +36,73 @@ def __repr__(self):
...
+        super(PolymorphClassSchema, self).__init__(
+            switch_field='path',
+            contents_map=contents_map,
+            description=description,
+            **kwargs
+        )
+
+        self.contents_map = self.DynamicContentsMap(enforce_object_type_subclass_of, self.contents_map)
+
+    class DynamicContentsMap(dict):
```
adamsussman: This feels like something you would want to use again.  Suggest making it a public class.

nickwilliams-eventbrite: GitHub PR comments are, frustratingly, not clear on the line on which you're actually commenting. To be clear, are you suggesting `DynamicContentsMap` be made a public class? If so, it doesn't really make sense to ever use this class on its own; it's necessarily very coupled with `PolymorphClassSchema`. With some considerable generifying, it could be extracted, but that seems like premature optimization absent a specific use case.

pysoa/common/schemas.py
```diff
@@ -34,3 +36,73 @@ def __repr__(self):
                 class_name=self.object_type.__name__,
             ) if self.object_type else '',
         )
+
+
+class PolymorphClassSchema(fields.Polymorph):
```
vartec: docstring please

pysoa/common/schemas.py
adamsussman: reasonable

pysoa/common/schemas.py
```diff
@@ -34,3 +36,73 @@ def __repr__(self):
...
+    def __init__(self, contents_map, enforce_object_type_subclass_of=None, description=None, **kwargs):
+        super(PolymorphClassSchema, self).__init__(
```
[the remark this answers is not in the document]
vartec: Yeah, for time being we have to keep Py2.7 compatibility :-/

pysoa/common/schemas.py
nickwilliams-eventbrite: Docstring added as requested.

PR #128, opened 2018-09-05
[PATCH] Allow extra keys to be censored via settings.

yanguo-eb:
Added sensitive field to redact in logging: `invitation_token`

### DIFF
--- a/pysoa/common/logging.py
+++ b/pysoa/common/logging.py
@@ -81,6 +81,7 @@ class RecursivelyCensoredDictWrapper(object):
         'auth', 'token', 'auth_token', 'auth-token', 'authToken',
         'authorization', 'authorization_token', 'authorization-token', 'authorizationToken',
         'authentication', 'authentication_token', 'authentication-token', 'authenticationToken',
+        'invitation_token',
 
         # Credit cards, banking, and related
         'credit_card', 'credit-card', 'creditCard', 'credit_card_number', 'credit-card-number', 'creditCardNumber',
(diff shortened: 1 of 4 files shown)


pysoa/common/logging.py
```diff
@@ -81,6 +81,7 @@ class RecursivelyCensoredDictWrapper(object):
         'auth', 'token', 'auth_token', 'auth-token', 'authToken',
         'authorization', 'authorization_token', 'authorization-token', 'authorizationToken',
         'authentication', 'authentication_token', 'authentication-token', 'authenticationToken',
+        'invitation_token',
```
vartec: no alternative spellings (`invitation`, `invitation-token`, `invitationToken`)?

yanguo-eb: Not foreseeing the variations ever being used in the context of pysoa.

renaudvisage:
I agree with you @nickwilliams-eventbrite.  I hesitated before accepting the PR but we definitely need a way to extend this list for service-specific purposes and this should be doable outside of pysoa.

setup.py
```diff
@@ -18,7 +18,7 @@ def readme():
 
 base_requirements = [
     'attrs~=17.4',
-    'conformity~=1.13',
+    'conformity~=1.70.0',
```
nickwilliams-eventbrite: 1.17.0, not 1.70.0

pysoa/common/settings.py
```diff
@@ -205,9 +205,15 @@ class SOASettings(Settings):
             description='The list of all middleware objects that should be applied to this server or client',
         ),
         'metrics': MetricsSchema(),
+        'extra_fields_to_redact': fields.Set(
+            fields.UnicodeString(),
+            description='Use this field to supplement the set of fields that are automatically redacted/censored in '
+                'request and response fields with additional fields that your service needs redacted.',
```
nickwilliams-eventbrite: Indent four more spaces.

setup.py
```diff
@@ -18,7 +18,7 @@ def readme():
 
 base_requirements = [
     'attrs~=17.4',
-    'conformity~=1.13',
+    'conformity~=1.17.0',
```
nickwilliams-eventbrite: Ah, man, I'm sorry. This was my bad. `'1.17.0'` is too specific for a library. Should be `'1.17'`.

PR #129, opened 2018-09-06
[MAJOR] Fix bug allowing missing `kwargs` in Redis, Local, and Stub transports

nickwilliams-eventbrite:
Adam pointed out a bug in our transport settings schemas that allowed users to omit `kwargs` entirely from the Redis, Local, and Stub transports (client and server) without failing validation, even though those transports won't work without those settings. This change:

- Replaces the `LocalTransportSchema` with `LocalClientTransportSchema` and `LocalServerTransportSchema` (a breaking change), as these actually have different requirements
- Removes `kwargs` from the list of optional keys in the Redis and Stub transport and the Local client transport (but not the Local server transport, which doesn't require `kwargs`)
- Regenerates documentation


sethbrite:
I know this is _technically_ a breaking change, but I'm not sure if this requires a major version. It's a fix for a bug you unintentionally introduced due to inheritance from `BasicClassSchema`.

nickwilliams-eventbrite:
>I know this is technically a breaking change, but I'm not sure if this requires a major version. It's a fix for a bug you unintentionally introduced due to inheritance from BasicClassSchema.

That doesn't matter so much, since we're still beta so we'll still increment the minor version instead of the major.

PR #134, opened 2018-10-01
[MINOR] Remove deprecated use of "encoding" argument in msgpack.unpackb

nickwilliams-eventbrite:
Since `msgpack-python` version 0.5.2, the `encoding` parameter is deprecated. `raw=False` is the replacement for `encoding='utf-8'`. This commit makes that change and bumps the library to require at least 0.5.2.

Related warning message:

```
/usr/local/lib/python3.5/dist-packages/ebsoa/server/middleware/auth.py:137: PendingDeprecationWarning: encoding is deprecated, Use raw=False instead.
```


cubes:
Looks good. 🚢

PR #135, opened 2018-10-03
[MINOR] Allow setting `side_effect` while defining the stub

bar:
This is currently possible only after the mock is returned by the context manager, which makes it impossible to define preconfigured stubs (specially when setting a callable that rely on variable data).

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -162,7 +162,7 @@ the real service is not configured).
 Calling ``stub_action``
 ***********************
 
-``stub_action`` has four potential arguments. Only the first two are required:
+``stub_action`` has five potential arguments. Only the first two are required:
 
 - ``service``: The name of the service on which this action will be called
 - ``action``: The name of the action to stub
@@ -170,17 +170,25 @@ Calling ``stub_action``
   would normally be returned from the action class's ``run`` method).
 - ``errors``: A list of SOA errors that should be raised, where each error is a dict with at least ``code`` and
   ``message`` keys and optionally a ``field`` for field errors.
+- ``side_effect``: A function, an exception class or instance, or an iterable. It behaves exactly like
+  ``mock.MagicMock.side_effect``.
 
 Instead of providing ``body`` and/or ``errors`` to ``stub_action``, you can manipulate the action stub object passed
 to the test method (or returned from the context manager) to tell it to return certain values or have certain side
 effects. The action stub object actually extends ``mock.MagicMock``, so you may already be very familiar with how it
 works.
 
-Given an action stub object ``stub_xx_action``, you can set ``stub_xx_action.return_value`` to control what
-the action returns (this is equivalent to the ``body`` argument to ``stub_action``). Alternatively, you can set
+Given an action stub object ``stub_xx_action``, you can set ``stub_xx_action.return_value`` to control what the action
+returns (this is equivalent to the ``body`` argument to ``stub_action``). Alternatively, you can set
 ``stub_xx_action.side_effect`` to raise SOA errors, provide different behavior for each of multiple expected calls, or
 exert more control over how the stub behaves. ``side_effect`` can be a single value of any of the following or a
 list/tuple (for multiple calls) where each value is any of the following:
+
+``side_effect`` is useful for raising exceptions or dynamically changing return values. The function is called with the
+same arguments as the mock, and unless it returns DEFAULT, the return value of this function is used as the return
+value. Alternatively ``side_effect`` can be an exception class or instance. In this case the exception will be raised
+when the mock is called. If ``side_effect`` is an iterable then each call to the mock will return the next value from
+the iterable.
 
 - A response body dict (same as the ``body`` argument to ``stub_action``)
 - An instance of ``ActionError`` with one or more SOA errors configured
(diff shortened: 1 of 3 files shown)


pysoa/test/stub_service.py
```diff
@@ -353,14 +353,15 @@ def call_body(self):
         def call_bodies(self):
             return tuple(args[0][0] for args in self.call_args_list)
 
-    def __init__(self, service, action, body=None, errors=None):
+    def __init__(self, service, action, body=None, errors=None, side_effect=None):
```
nickwilliams-eventbrite: I think I would like to call this `function` instead of `side_effect`. Given that this is a stub (even if it uses a Mock to do the heavy lifting), I feel like `function` more clearly portrays its purpose. What do you think? I'm not hard set on that.

pysoa/test/stub_service.py
```diff
@@ -378,7 +379,10 @@ def __enter__(self):
         mock_action = self._MockAction(name='{}.{}'.format(self.service, self.action))
 
         if self.body or self.errors:
-            mock_action.side_effect = lambda _: ActionResponse(self.action, errors=self.errors, body=self.body)
+            mock_action.return_value = ActionResponse(self.action, errors=self.errors, body=self.body)
+
+        if self.side_effect:
```
nickwilliams-eventbrite: This needs to be `elif`.

pysoa/test/stub_service.py
```diff
@@ -353,14 +353,15 @@ def call_body(self):
         def call_bodies(self):
             return tuple(args[0][0] for args in self.call_args_list)
 
-    def __init__(self, service, action, body=None, errors=None):
+    def __init__(self, service, action, body=None, errors=None, side_effect=None):
         assert isinstance(service, six.text_type), 'Stubbed service name "{}" must be unicode'.format(service)
         assert isinstance(action, six.text_type), 'Stubbed action name "{}" must be unicode'.format(action)
```
nickwilliams-eventbrite: You need an `assert not (body or actions) or not side_effect`. It's not valid to provide both.

pysoa/test/stub_service.py
bar: I'd rather stick to `side_effect`:
```
...

Alternatively side_effect can be an exception class or instance. In this case the exception will be raised when the mock is called.

If side_effect is an iterable then each call to the mock will return the next value from the iterable.
```

pysoa/test/stub_service.py
bar: you should be able to define both, `mock` deals with it internally, useful when you want the callable to do something interesting, but at the same time you want to set a `return_value`

pysoa/test/stub_service.py
nickwilliams-eventbrite: Okay.

pysoa/test/stub_service.py
nickwilliams-eventbrite: Hmmm. Okay. I can live with that. But I think the documentation should explicitly state that it's not recommended (can be confusing to people who later come and edit tests using this feature) and should only be used if you really know what you're doing.

docs/testing.rst
```diff
@@ -162,26 +162,34 @@ the real service is not configured).
...
-``stub_action`` has four potential arguments. Only the first two are required:
+``stub_action`` has five potential arguments. Only the first two are required:
 
 - ``service``: The name of the service on which this action will be called
 - ``action``: The name of the action to stub
 - ``body``: A dictionary containing the response body that the stubbed action should return (the same schema that
   would normally be returned from the action class's ``run`` method).
 - ``errors``: A list of SOA errors that should be raised, where each error is a dict with at least ``code`` and
   ``message`` keys and optionally a ``field`` for field errors.
+- ``side_effect``: A function, an exception class or instance, or an iterable. It behaves exactly like
```
nickwilliams-eventbrite: Please add:

>While it's possible to use `body` or `errors` in conjunction with `side_effect`, it is not recommended unless you really know what you're doing. It can be confusing for future developers working in your code. See the Python `unittest.Mock` documentation for a detailed description of the expected behavior when specifying `return_value` and `side_effect` at the same time.

docs/testing.rst
```diff
@@ -162,26 +162,34 @@ the real service is not configured).
...
 
-Given an action stub object ``stub_xx_action``, you can set ``stub_xx_action.return_value`` to control what
-the action returns (this is equivalent to the ``body`` argument to ``stub_action``). Alternatively, you can set
+Given an action stub object ``stub_xx_action``, you can set ``stub_xx_action.return_value`` to control what the action
+returns (this is equivalent to the ``body`` argument to ``stub_action``). Alternatively, you can set
 ``stub_xx_action.side_effect`` to raise SOA errors, provide different behavior for each of multiple expected calls, or
 exert more control over how the stub behaves. ``side_effect`` can be a single value of any of the following or a
 list/tuple (for multiple calls) where each value is any of the following:
 
+``side_effect`` is useful for raising exceptions or dynamically changing return values. The function is called with the
```
nickwilliams-eventbrite: This new paragraph is in the wrong place. It interrupts "any of the following:" with the list (below) to which "following" refers. Please move this paragraph down below the list and change "`side_effect` is useful" to "The `side_effect` argument is useful."

pysoa/test/stub_service.py
yanguo-eb: Is it a breaking change to any existing tests using `side_effect`?

bar: nope, cause they are both (old and new) evaluated before the mock is created and returned to the test itself

PR #138, opened 2018-10-11
[MINOR] A better Syslog logging handler

nickwilliams-eventbrite:
The Syslog logging handler built into Python has a crucial flaw: If the total, formatted log message exceeds the minimum interface MTU between the service and the Syslog server, that entire message is lost if sent over UDP (the default and best-performing option). It will either log the error or silently fail, but the message will be lost. Since PySOA logging includes request/response logging, which can be large, it means we can possibly lose a lot of logged information that we didn't expect to lose, without even know we lost it.

This commit provides an improvement on that built-in Syslog handler. At startup, it will attempt to determine the minimum MTU between the running process and the Syslog server, or make an educated guess if it can't (or set an upper boundary of 1MB over TCP and Unix sockets). It will then either truncate all messages to come in under that MTU or fragment messages into multiple Syslog messages, based on the configured setting.

This Syslog handler replaces the built-in Syslog handler in the default Server settings.

### DIFF
--- a/pysoa/common/logging.py
+++ b/pysoa/common/logging.py
@@ -4,6 +4,8 @@ from __future__ import (
 )
 
 import logging
+import logging.handlers
+import socket
 import threading
 
 import six
@@ -176,6 +178,7 @@ class RecursivelyCensoredDictWrapper(object):
 
     def __unicode__(self):
         # If this method is called, we must be in Python 2, which means __str__ must be returning bytes
+        # noinspection PyUnresolvedReferences
         return self.__str__().decode('utf-8')
 
     @classmethod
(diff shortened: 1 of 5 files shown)


pysoa/common/logging.py
```diff
@@ -199,3 +202,179 @@ def _copy_and_censor_dict(cls, d):
...
+                # contains "too long" if we're getting an "MTU exceeded" ICMP response. IF this isn't an "MTU exceeded"
+                # response, then something went wrong, and we can't determine the MTU, so fall back to worst case.
+                return WORST_CASE_MTU_IP
+            # Now we try again with a lower MTU, which might still be too high
+        else:
+            return attempt
+    return WORST_CASE_MTU_IP
+
+
+class SyslogHandler(logging.handlers.SysLogHandler):
```
bar: Please rename to `SysLogHandler` (not the capital `L`) to be consistent with `logging.handlers.SysLogHandler` and config issues... this is very likely to get mistyped otherwise

pysoa/common/logging.py
```diff
@@ -199,3 +202,179 @@ def _copy_and_censor_dict(cls, d):
...
+    s.setsockopt(socket.IPPROTO_IP, IP_MTU_DISCOVER, IP_MTU_DISCOVER_DO)
+    for attempt in MTU_ATTEMPTS:
+        try:
+            s.send(b'#' * (attempt - DATAGRAM_HEADER_LENGTH_IN_BYTES))
+        except socket.error as e:
+            if 'too long' not in e.strerror:
+                # We can't rely on the error code, because it varies from platform to platform, but the message always
+                # contains "too long" if we're getting an "MTU exceeded" ICMP response. IF this isn't an "MTU exceeded"
+                # response, then something went wrong, and we can't determine the MTU, so fall back to worst case.
+                return WORST_CASE_MTU_IP
```
bar: I don't see the point of `if 'too long' not in e.strerror:` if it does indeed hit it, it will return the same as if it didn't

`else` is only triggered if no exception is raised

which is equivalent to
```
    for attempt in MTU_ATTEMPTS:
        try:
            s.send(b'#' * (attempt - DATAGRAM_HEADER_LENGTH_IN_BYTES))
            return attempt
        except socket.error as e:
            if 'too long' not in e.strerror:
                # We can't rely on the error code, because it varies from platform to platform, but the message always
                # contains "too long" if we're getting an "MTU exceeded" ICMP response. IF this isn't an "MTU exceeded"
                # response, then something went wrong, and we can't determine the MTU, so fall back to worst case.
                pass
    return WORST_CASE_MTU_IP
```

pysoa/common/logging.py
nickwilliams-eventbrite: I disagree with this. Point of fact: The handler name in Python is incorrect (just like all the time zone stuff in Python is misnamed and should be `time_zone`/`TimeZone` instead of `timezone`/`Timezone`). The proper name is Syslog, not SysLog, so the correct name for this handler is `SyslogHandler`, not `SysLogHandler`. I'd rather have a correct name that doesn't match Python than an incorrect name that does match Python.

pysoa/common/logging.py
nickwilliams-eventbrite: I agree it can be simplified, but only marginally. The point of `if 'too long' not in e.strerror:` is that, if we get that error, then something else has gone wrong (perhaps an invalid address or broken network, for example), and we need to just stop looping immediately and return the worst-case. We're likely to just keep getting that error over and over again with every attempt before eventually getting to the worst case. The only error that is acceptable for continuing is the "too long" error. I can, however, get rid of the `else`, move the `return attempt`, and change one `return WORST_CASE_MTU_IP` to a `break`.

```
def _discover_minimum_mtu_to_target(address, port):
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    s.connect((address, port))
    s.setsockopt(socket.IPPROTO_IP, IP_MTU_DISCOVER, IP_MTU_DISCOVER_DO)
    for attempt in MTU_ATTEMPTS:
        try:
            s.send(b'#' * (attempt - DATAGRAM_HEADER_LENGTH_IN_BYTES))
            return attempt
        except socket.error as e:
            if 'too long' not in e.strerror:
                # We can't rely on the error code, because it varies from platform to platform, but the message always
                # contains "too long" if we're getting an "MTU exceeded" ICMP response. IF this isn't an "MTU exceeded"
                # response, then something went wrong, and we can't determine the MTU, so fall back to worst case.
                break
            # Now we try again with a lower MTU, which might still be too high
    return WORST_CASE_MTU_IP
```

pysoa/common/logging.py
```diff
@@ -199,3 +202,178 @@ def _copy_and_censor_dict(cls, d):
...
+                self.socket.sendto(message, self.address)
+            else:
+                self.socket.sendall(message)
+
+    @staticmethod
+    def _cleanly_slice_encoded_string(encoded_string, length_limit):
+        sliced, remaining = encoded_string[:length_limit], encoded_string[length_limit:]
+        try:
+            sliced.decode('utf-8')
+        except UnicodeDecodeError as e:
```
vartec: OK, so I figures out why this will _always_ work correctly, but a comment in the code would be nice. It works, because the first byte of a UTF-8 character encodes how many bytes are in total in that character.

nickwilliams-eventbrite: Docstring added.

PR #140, opened 2018-10-22
Added django exception to ImproperlyConfigured when loading server

viniciusfeitosa:
When PySOA is loading the server, Django is on the PYTHON_PATH, and the DJANGO_SETTINGS_MODULE isn't configured properly a Django exception ImproperlyConfigured have been raising. This PR fix it.


viniciusfeitosa:
This PR isn't necessary.

PR #143, opened 2018-10-26
[MINOR] Prevent server shutdown on request with non-unicode context keys

adamsussman:
Server would shut down entirely if a request was received with non unicode keys in the context.  It appears to be a python issue when unpacking a dict with non-unicode keys into function arguments.

Not sure if something should be logged here.

The exception in this spot caused pysoa to choose to shutdown with:

2018-10-26 15:46:13,786   ERROR -- --: Unhandled server error; shutting down
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/pysoa/server/server.py", line 626, in run
    self.handle_next_request()
  File "/usr/local/lib/python3.5/dist-packages/pysoa/server/server.py", line 177, in handle_next_request
    PySOALogContextFilter.set_logging_request_context(request_id=request_id, **job_request['context'])
TypeError: set_logging_request_context() keywords must be strings
2018-10-26 15:46:13,787    INFO -- --: Server shutting down
2018-10-26 15:46:13,787    INFO -- --: Stopping and closing async event loop
2018-10-26 15:46:13,787    INFO -- --: Closing all Django caches
2018-10-26 15:46:13,787    INFO -- --: Closing and removing heartbeat file

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -1009,4 +1009,5 @@ class Client(object):
         # Add any extra stuff
         if context_extra:
             context.update(context_extra)
-        return context
+        # context keys need to be guaranteed unicode
+        return {six.text_type(k): v for k, v in six.iteritems(context)}
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -1009,5 +1009,4 @@ class Client(object):
         # Add any extra stuff
         if context_extra:
             context.update(context_extra)
-        # context keys need to be guaranteed unicode
-        return {six.text_type(k): v for k, v in six.iteritems(context)}
+        return context
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -177,8 +177,12 @@ class Server(object):
         try:
             PySOALogContextFilter.set_logging_request_context(request_id=request_id, **job_request['context'])
         except TypeError:
-            # non unicode keys in job_request['context'] will break keywording of a function call
-            pass
+            # Non unicode keys in job_request['context'] will break keywording of a function call.
+            # Try to recover by coercing the keys
+            PySOALogContextFilter.set_logging_request_context(
+                request_id=request_id,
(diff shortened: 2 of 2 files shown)


pysoa/server/server.py
```diff
@@ -174,7 +174,11 @@ def handle_next_request(self):
         self._idle_timer.stop()
         self._idle_timer = None
 
-        PySOALogContextFilter.set_logging_request_context(request_id=request_id, **job_request['context'])
+        try:
+            PySOALogContextFilter.set_logging_request_context(request_id=request_id, **job_request['context'])
+        except TypeError:
+            # non unicode keys in job_request['context'] will break keywording of a function call
+            pass
```
nickwilliams-eventbrite: Don't pass. Please do this:

```
except TypeError:
    PySOALogContextFilter.set_logging_request_context(
        request_id=request_id, 
        **(six.text_type(k): v for k, v in six.iteritems(job_request['context'])),
    )
```

I would greatly prefer not to lose this logging information.

pysoa/client/client.py
```diff
@@ -1009,4 +1009,5 @@ def _make_context_header(self, switches=None, correlation_id=None, context_extra
         # Add any extra stuff
         if context_extra:
             context.update(context_extra)
-        return context
+        # context keys need to be guaranteed unicode
+        return {six.text_type(k): v for k, v in six.iteritems(context)}
```
nickwilliams-eventbrite: I have concerns about this. This is going to add overhead to every PySOA request, whether it needs it or not (whereas the below only adds the overhead when needed). In case there was any misunderstanding, when I said this:

>please also fix whatever caused your client to send bytes objects as context dict keys

I wasn't referring to making a change to the PySOA client code. I was referring to fixing whatever code was using the client and sending non-unicode keys to it.

Perhaps I'm splitting hairs here and the overhead doesn't really matter. I'm not outright rejecting it, just saying that I have concerns. 🤔

PR #145, opened 2018-11-19
[PATCH] Throttle updates of the heartbeat file

nickwilliams-eventbrite:
We don't want to update the heartbeat file every request. We could tank the file system and kill performance. At most, it should be updated every 2.5 seconds.

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -150,6 +150,7 @@ class Server(object):
 
         self._heartbeat_file = None
         self._heartbeat_file_path = None
+        self._heartbeat_file_last_update = 0
 
     def handle_next_request(self):
         """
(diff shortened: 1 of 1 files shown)


pysoa/server/server.py
```diff
@@ -557,10 +558,13 @@ def _delete_heartbeat_file(self):
                     self.logger.exception('Error while removing heartbeat file')
 
     def _update_heartbeat_file(self):
-        if self._heartbeat_file:
+        if self._heartbeat_file and time.time() - self._heartbeat_file_last_update > 2.5:
+            # Only update the heartbeat file if one is configured and it has been at least one second since last update.
```
sethbrite: One second or 2.5 seconds?

nickwilliams-eventbrite: Dope! Comment updated.

PR #148, opened 2019-01-29
[PATCH] Use client timeout for expansions receive responses

renzo-eb:
Use the same timeout specified in the client for the main call
to wait for expansions responses.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -951,7 +951,10 @@ class Client(object):
             for service_name, request_ids_to_objects in expansion_service_requests.items():
                 if request_ids_to_objects:
                     # Receive all available responses from the service
-                    for request_id, response in self.get_all_responses(service_name):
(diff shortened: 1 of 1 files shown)


nickwilliams-eventbrite:
In my opinion, you're taking too complicated of an approach to this. I don't see, at this time, a strong need to provide differentiated per-service timeouts for expansions. Remember that you're committing to a public interface when you add this, because this is OSS. Until a need has been demonstrated for such a complex feature, I think we should err toward a more conservative approach.

I think what it boils down to is that there is a bug. That bug is that the `timeout` argument passed to `call_action*****` is not passed down through `_perform_expansion` and `_expand_objects` to the expansion service calls like it should be. So just change the code to make sure that `timeout` gets passed on to expansions. That's simpler, and it's easy to still add a more complex feature later if the need arises.

renzo-eb:
Updated @nickwilliams-eventbrite , thanks

PR #150, opened 2019-02-14
[PATCH] Run the event loop in a separate thread.

sethbrite:
The PySOA server currently creates an `asyncio` event loop (if `asyncio` is available) and makes it available to jobs/actions. However, this event loop is never started and will never run scheduled tasks and coroutines.

This change creates and runs an `asyncio` event loop in a separate thread. This allows users to run async coroutines/tasks in the background.

NOTE: all coroutines/callbacks scheduled against this event loop must use the `*_threadsafe` methods. See the documentation here: https://docs.python.org/3/library/asyncio-dev.html#concurrency-and-multithreading

### DIFF
--- a/pysoa/server/internal/event_loop.py
+++ b/pysoa/server/internal/event_loop.py
@@ -2,8 +2,7 @@ import asyncio
 import threading
 
 
-def _loop_stop_callback():
-    loop = asyncio.get_running_loop()
+def _loop_stop_callback(loop):
     loop.stop()
 
 
@@ -20,7 +19,7 @@ class AsyncEventLoopThread(threading.Thread):
             self.loop.run_forever()
         finally:
             try:
-                pending_tasks = asyncio.all_tasks(self.loop)
+                pending_tasks = asyncio.Task.all_tasks(self.loop)
                 self.loop.run_until_complete(asyncio.gather(*pending_tasks))
             finally:
                 self.loop.close()
@@ -29,7 +28,7 @@ class AsyncEventLoopThread(threading.Thread):
 
     def join(self):
         if self.loop.is_running():
-            self.loop.call_soon_threadsafe(_loop_stop_callback)
+            self.loop.call_soon_threadsafe(_loop_stop_callback, self.loop)
         self._done.wait()
 
     def run_coroutine(self, coroutine):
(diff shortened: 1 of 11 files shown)


pysoa/server/server.py
```diff
@@ -53,9 +53,10 @@
 
 
 try:
-    import asyncio
+    from pysoa.server.internal.event_loop import AsyncEventLoopThread
+    has_asyncio = True
 except ImportError:
-    asyncio = None
+    has_asyncio = False
```
nickwilliams-eventbrite: Why not just `AsyncEventLoopThread = None`? I ask because my IDE warns me about this, says that `AsyncEventLoopThread` is undefined, but setting `AsyncEventLoopThread = None` here clears that warning and also obviates any need for `has_asyncio`.

pysoa/server/internal/event_loop.py
```diff
@@ -0,0 +1,35 @@
+import asyncio
+import threading
+
+
+def _loop_stop_callback(loop):
```
nickwilliams-eventbrite: Can you make this a `@staticmethod` on the `AsyncEventLoopThread` class instead of making it a function in the module? Would you even need the argument then, or could you just use `self.loop.stop()`?

pysoa/server/server.py
```diff
@@ -107,8 +108,9 @@ def __init__(self, settings):
             raise AttributeError('Server subclass must set service_name')
 
         self.async_event_loop = None
-        if asyncio:
-            self.async_event_loop = asyncio.get_event_loop()
+        if has_asyncio:
+            self._async_event_loop_thread = AsyncEventLoopThread()
+            self.async_event_loop = self._async_event_loop_thread.loop
```
nickwilliams-eventbrite: A couple things:
- You can make `async_event_loop` private (`_async_event_loop`)
- We're currently adding the `async_event_loop` to the job request (https://github.com/eventbrite/pysoa/blob/e195507f3673903744abd5675018324b198c3aad/pysoa/server/server.py#L313) and to the action request (https://github.com/eventbrite/pysoa/blob/e195507f3673903744abd5675018324b198c3aad/pysoa/server/server.py#L413). Since we want everybody to use `AsyncEventLoopThread.run_coroutine`, we should transition away from adding the `async_event_loop` attribute to the job/action requests and instead add an `run_async_coroutine` (a lambda that calls `AsyncEventLoopThread.run_coroutine`, I don't want to expose the whole thread where someone could call `run` or `join`). We might have to have them both at the same time as a transition period.

PR #151, opened 2019-02-19
[PATCH] Fix tests broken by latest PyTest version

nickwilliams-eventbrite:
PyTest 4.2.0 introduced a regression which broke the tests in our test plans, but wasn't a deal-breaker, so we simply adapted to the breakage. The regression was fixed in 4.2.1, but the fix introduced a new feature (not a bug) that `@unittest.skip`-skipped classes won't even be handled by our test plan code, so we can't measure them in the plugin-testing tests. This commit adapts to that change and removes the adaptation to the fixed regression. At also adds some new assertions and information to help ensure this is still working, even with the new changes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ test_helper_requirements = [
 
 test_plan_requirements = test_helper_requirements + [
     'pyparsing~=2.2',
-    'pytest>=3.1,<5',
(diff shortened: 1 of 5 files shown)


ebwill:
looks good to me

PR #153, opened 2019-03-13
[MAJOR] Step 2 in the message serializer content type header

nickwilliams-eventbrite:
This is the second step (commit) in a three-step (three-commit) breaking change to add support for switching the message serializer that the Redis transport uses based on a content-type header. There will be one more breaking commit in a future version, in order to make migration possible.

Breaking changes in this commit/step:
- All request and response messages the Redis transport sends now contain a leading content-type header containing the MIME type used for selecting the proper deserializer on the receiving end. Before you upgrade to PySOA version 0.58.0 and get this change, you must first upgrade all clients and servers to at least 0.55.0 but still less than 0.58.0; otherwise, older clients or servers might not understand messages they receive from clients and servers running 0.58.0 or newer.
- (Unrelated) A previously-announced breaking change has been finalized: `check_*` methods in actions extending the `BaseStatusAction` must now accept exactly one parameter, which will be the `ActionRequest`.

Explanation of the steps:
- Step 1 (1cc1d28787d630fc3973497911a0e6ecf4867cd4): Rename the configuration setting, make the `receive_message` method use a switched serializer if the content type header is present, and make the `send_message` method use a switched serializer and include a content type header if the `receive_message` method found a content type header. Deploy this to all servers and clients.
- Step 2 (this commit): Make the `send_message` method always send a content type header. Deploy this to all servers and clients.
- Step 3: Make the `receive_message` method enforce the presence of a content type header. Deploy this to all servers and clients.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -182,4 +182,21 @@ have the service code in place, and a ``stub_action`` decorator / context manage
 For more information about using these test utilities in your services or service-calling applications, see the testing
 documentation in the `PySOA Documentation <docs/index.rst>`_.
 
-For testing this PySOA library directly on your system, you must first install `Docker <https://www.docker.com/get-started>`_. Once installed, the PySOA test image can be built with the following command: ``docker build . -t pysoa-test``. Once the pysoa-test image has been built, run ``docker run pysoa-test`` to execute the test suite. The test suite will run for all supported versions of Python.
+For testing this PySOA library directly on your system, you must first install `Docker
+<https://www.docker.com/get-started>`_. One installed, you can run tests across all supported environments using one
+or more of the following commands::
+
+    # Run all tests in Python 2.7, 3.5, 3.6, and 3.7, do Flake8 analysis, and do code coverage analysis
+    ./tox.sh
+
+    # Run all tests in Python 3.5
+    ./tox.sh -e py35
+
+    # Run all tests in Python 2.7 and 3.7
+    ./tox.sh -e py27,py37
+
+    # Run all tests in Python 3.5, 3.6, and 3.7 and do code coverage analysis
+    ./tox.sh -e py35,py36,py37,coverage
+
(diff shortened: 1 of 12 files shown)


README.rst
```diff
@@ -182,4 +182,21 @@ have the service code in place, and a ``stub_action`` decorator / context manage
...
+or more of the following commands::
+
+    # Run all tests in Python 2.7, 3.5, 3.6, and 3.7, do Flake8 analysis, and do code coverage analysis
+    ./tox.sh
+
+    # Run all tests in Python 3.5
+    ./tox.sh -e py35
+
+    # Run all tests in Python 2.7 and 3.7
+    ./tox.sh -e py27,py35
```
miloops: The comment doesn't match the command, 3.7 and py35, please change any of them to match

nickwilliams-eventbrite: Good catch! Fixed.

tests/common/test_logging.py
```diff
@@ -290,23 +291,62 @@ def test_complex_dict(self):
...
+        else:
+            assert handler.unixsocket is False
         assert handler.overflow == SyslogHandler.OVERFLOW_BEHAVIOR_TRUNCATE
         assert handler.maximum_length >= 1252  # (1280 - 28)
 
         with mock.patch.object(socket.socket, 'connect'):
             handler = SyslogHandler(socket_type=socket.SOCK_STREAM)
             assert handler.socktype == socket.SOCK_STREAM
-            assert handler.unixsocket is False
+            if six.PY2:
```
vartec: why not specifically `in sys.version_info < (2, 7, 14)`?

nickwilliams-eventbrite: Meh. Easier to just use `six.PY2`. Your suggestion is a new import plus probably an explanation about why such a specific version number. It's going away eventually, anyway.

PR #158, opened 2019-04-15
[PATCH] Use Tox to add tests for PyInotify

nickwilliams-eventbrite:
Added another Tox config to run tests with PyInotify installed. Added additional tests that run only when PyInotify is installed. New tests pass without any changes to code.

### DIFF
--- a/tests/server/test_autoreloader.py
+++ b/tests/server/test_autoreloader.py
@@ -3,10 +3,12 @@ from __future__ import (
     unicode_literals,
 )
 
+import codecs
 import importlib
 import os
 import signal
 import sys
+import tempfile
 import time
 import unittest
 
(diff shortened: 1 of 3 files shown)


tests/server/test_autoreloader.py
```diff
@@ -494,45 +528,119 @@ def test_main_some_args(self):
...
+                mock_get_watch_file_names.return_value = [file1.name, file2.name]
 
-            self.assertEqual(2, mock_os.stat.call_count)
-            self.assertEqual('/path/to/first/file.py', mock_os.stat.call_args_list[0][0][0])
-            self.assertEqual('/path/to/another/level.py', mock_os.stat.call_args_list[1][0][0])
+                self.assertFalse(reloader.code_changed())
 
-            mock_os.reset_mock()
-            mock_os.stat.side_effect = [self.Stat(123456789, 0), self.Stat(987654321, 0)]
+                time.sleep(1.1)
```
vartec: :-/

nickwilliams-eventbrite: Yeah, I wanted to test with actual temporary files instead of mocking `os.stat`. However, `st_mtime` is measured in seconds, so the only way to do this is to sleep.

PR #166, opened 2019-05-23
add missing field to EnrichedActionRequest

eb-francisco:
add `metrics` as an input to `EnrichedActionRequest`


nickwilliams-eventbrite:
Discussed offline on Slack.

PR #167, opened 2019-05-24
[MINOR] Add helper for calling local actions within other actions

nickwilliams-eventbrite:
Add a helper method to `EnrichedActionRequest` to make it easy to call other local actions within the same service. Also, make `EnrichedActionRequest` subclassable by adding a `request_class` attribute to the `Server` class.

### DIFF
--- a/pysoa/server/types.py
+++ b/pysoa/server/types.py
@@ -3,19 +3,39 @@ from __future__ import (
     unicode_literals,
 )
 
-import attr
+from typing import (  # noqa: F401 TODO Python 3
+    Callable,
+    Dict,
+)
 
-from pysoa.common.types import ActionRequest
+import attr
+import six  # noqa: F401 TODO Python 3
+
+from pysoa.common.constants import (
+    ERROR_CODE_SERVER_ERROR,
+    ERROR_CODE_UNKNOWN,
+)
+from pysoa.common.types import (
+    ActionRequest,
+    ActionResponse,
+    Error,
+)
+from pysoa.server.errors import ActionError
 from pysoa.server.internal.types import RequestSwitchSet
 
 
 @attr.s
 class EnrichedActionRequest(ActionRequest):
     """
-    The action request object that the Server passes to each Action class that it calls.
+    The action request object that the Server passes to each Action class that it calls. It contains all the information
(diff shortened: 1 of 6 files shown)


pysoa/server/types.py
```diff
@@ -24,5 +44,68 @@ class EnrichedActionRequest(ActionRequest):
...
+        request = self.__class__(
+            action=action,
+            body=body,
+            # Dynamically copy all Attrs attributes so that subclasses introducing other Attrs can still work properly
+            **{a.name: getattr(self, a.name) for a in self.__attrs_attrs__ if a.name not in ('action', 'body')}
+        )
+        request._server = server
+
+        try:
+            response = action_callable(request)
```
adamsussman: Should middleware be run here?  What should users of this expect if they are used to, for example, a metrics middleware to be running on their actions normally?  It would be good to have clearer documentation on the differences (if any) in how this local action would run from an externally triggered action and what environmental expectations are appropriate.

nickwilliams-eventbrite: Great question, @adamsussman. I actually looked at doing this at first. There are some complications involved, and I would need to proceed carefully. I'm not sure I know the exact right way to do it yet. Critically, I looked through all of our internal middleware and I don't yet see any use cases where running the middleware here would make something work that otherwise wouldn't work, except maybe that we wouldn't record metrics around this action (which might even be desirable) (but the metrics container will still be passed in to the action, so it can still record _custom_ metrics). I'm not ruling out adding middleware in here in the future, but I'm not sure it's necessary _right now_. However, if you believe it is necessary, I can spend some more time on this now instead of putting that off.

Keep in mind that the way we are currently doing this (manually instantiating the class and calling the action) also does not run any middleware, so we aren't taking a step back from that. This is actually an improvement over our current process. Keep in mind, also, that there are other things we are skipping here: server metrics, request logging, etc. We're not treating this as a request. It's basically a helper to locally use the action class directly. For whatever that is worth.

adamsussman: I'm on the fence on whether its necessary.  If nothing else though, clear documentation on exactly what is or is not happening would be good.

nickwilliams-eventbrite: Fair enough. I have expanded the documentation to specifically explain the things that do not happen and clarify that we may add middleware execution to the helper in the future.

PR #171, opened 2019-05-31
[PATCH] Return same stub in multiple uses of the same stub_action instance

michaelmanganiello-eb:
PR #170 makes multiple uses of the same `stub_action()` instance to
have different return values on each call. This fixes it, so they have
return the same instead.

### DIFF
--- a/pysoa/test/stub_service.py
+++ b/pysoa/test/stub_service.py
@@ -368,24 +368,25 @@ class stub_action(object):  # noqa
         self.attribute_name = None
         self.new = mock.DEFAULT
 
+        self._current_mock_action = None
         self._stub_action_responses_outstanding = defaultdict(dict)
         self._stub_action_responses_to_merge = defaultdict(dict)
 
     def __enter__(self):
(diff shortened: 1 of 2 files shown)


pysoa/test/stub_service.py
```diff
@@ -372,20 +372,20 @@ def __init__(self, service, action, body=None, errors=None, side_effect=None):
         self._stub_action_responses_to_merge = defaultdict(dict)
 
     def __enter__(self):
-        mock_action = self._MockAction(name='{}.{}'.format(self.service, self.action))
-
         if self.enabled:
-            return mock_action
+            return self.mock_action
+
+        self.mock_action = self._MockAction(name='{}.{}'.format(self.service, self.action))
```
nickwilliams-eventbrite: Can we call this `self._current_mock_action`, instead, please? Also, we should initialize `self._current_mock_action = None` in the constructor.

michaelmanganiello-eb: Fixed, thanks!

PR #176, opened 2019-07-05
[MINOR] Support PyTest 5.0 with tests ensuring compliance

nickwilliams-eventbrite:


pysoa/test/plugins/pytest/plans.py
```diff
@@ -130,6 +130,9 @@ def pytest_pycollect_makeitem(collector, name, obj):
     if not TEST_PLANS_ENABLED:
         return
 
+    if not isinstance(obj, type):
```
nickwilliams-eventbrite: This wasn't necessary to support 5.0. I placed it here while troubleshooting an issue I was having, and it seemed like a worthy thing to leave in place.

PR #181, opened 2019-07-29
[PATCH] Fix documentation introspection in Python 2

apizarro-eb:
Fix an issue with `IntrospectionAction` that results in a response validation error in
Python 2 when an action doesn't have a unicode description or docstring.

### DIFF
--- a/pysoa/server/action/introspection.py
+++ b/pysoa/server/action/introspection.py
@@ -221,8 +221,10 @@ class IntrospectionAction(Action):
 
     @staticmethod
     def _introspect_action(action_class):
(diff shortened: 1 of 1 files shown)


nickwilliams-eventbrite:
Hmm. I'm not sure about this change. In a PySOA service, all text strings must always be unicode strings. Byte strings are only permitted when transacting non-text binary data (such as encoded/encrypted stuff). Having a `description` or `__doc__` that is not a unicode string is not valid in PySOA, so I'm not inclined to accept a change that would permit it. What do you think, @sethbrite?

PR #187, opened 2019-09-03
[MAJOR] Support UTC-aware datetime objects in MsgpackSerializer

nickwilliams-eventbrite:
`MsgpackSerializer` now supports both naive and UTC-aware `datetime.datetime` objects. Other time zones, however, are not supported. This is a breaking change, because sending a UTC-aware `datetime` from a client to a server on an older version will cause deserialization errors and you will get an error response. Be sure that all servers are running this version of PySOA before sending requests contaiting UTC-aware `datetime` objects.

### DIFF
--- a/pysoa/common/serializer/msgpack_serializer.py
+++ b/pysoa/common/serializer/msgpack_serializer.py
@@ -12,6 +12,7 @@ import struct
 from conformity import fields
 import currint
 import msgpack
+import pytz
 import six
 
 from pysoa.common.serializer.base import Serializer as BaseSerializer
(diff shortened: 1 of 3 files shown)


pysoa/common/serializer/msgpack_serializer.py
```diff
@@ -67,6 +68,7 @@ class MsgpackSerializer(BaseSerializer):
     EXT_CURRINT = 2
     EXT_DATE = 3
     EXT_DATETIME = 1
+    EXT_DATETIME_UTC = 10
```
bar: `10` looks a deliberately chosen, when the next logical id would be `6`, why?

bar: @nickwilliams-eventbrite why are we assuming extensions will always be less than 9 (aka be represented by 1 digit)?

PR #189, opened 2019-09-05
[MAJOR] Add support for response chunking to Redis Gateway transport

nickwilliams-eventbrite:
This implements #190 and part of #195.

- Part of the 1.0.0 milestone includes adding Python Typing support throughout the library. That was not the goal of this commit, but doing so in part of the library made it easier to correctly implement these changes, so this commit includes extensive added Python Typing support.
- Add a new Protocol Version concept to the PySOA Redis transport protocol.
- Although Protocol Version was previously undocumented, it was still an implicit concept, where without-content-type was Version 1 and with-content-type was Version 2. Now that it is a formal concept, the default Protocol Version has changed from Version 1 to Version 2, with a client-side configuration option to override the default with Version 1, 2, or 3.
- Add server response chunking to the PySOA Redis Transport Protocol version 3. Client requests can never support chunking, due to the nature of the way the protocol works at its foundation. Only server responses can support chunking.
- Add a bunch of unit tests to verify the chunking implementation.
- Add more functional tests to verify the chunking behavior, and found and fixed some bugs as a result.
- Update documentation.
- Make auto-doc-generation work in Python 3 (only).
- Re-run auto-doc-generation.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -821,6 +821,15 @@ The Redis Gateway transport takes the following extra keyword arguments for conf
   messages larger than this (set this to 0 to disable the warning)
 - ``maximum_message_size_in_bytes``: Defaults to 102,400 bytes on the client and 256,000 bytes on the server, defines
   the threshold at which ``MessageTooLarge`` will be raised.
+- ``chunk_messages_larger_than_bytes``: This option exists only for the Server transport and not for the Client
+  transport and controls the threshold at which responses will be chunked. Chunked responses allows your servers to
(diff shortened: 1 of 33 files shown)


nickwilliams-eventbrite:
> Side note: can we have one PR per feature in the future, diffs as large as this one, with unrelated features are annoying to review.

That was originally the goal, but once I got part-way through adding chunking support, I realized I was adding typing information to help me feel more confident in the changes, and then I realized I needed //more// typing information to continue feeling confident in the changes, and by that point I had two parallel changes (chunking and type comments) that were suddenly difficult to separate. 🤷‍♂

PR #211, opened 2019-10-05
[MINOR] Refactor PySOA errors for easier and more concise usage

nickwilliams-eventbrite:
- Create a common base class for all PySOA errors, `PySOAError`, and ensure all publicly-exposed exceptions inherit from that.
- Create common base classes `PySOAClientError`, `PySOAServerError`, `PySOATransportError`, `TransientPySOATransportError`, and `SerializationError` for more concise (less verbose) catching of related errors.
- The library had a mixture of `errors` modules and `exceptions` modules; standardize on `errors` and deprecated the `exceptions` modules.
- `is_caller_error` is brand new and already the `is_caller_error` arguments to `ActionError` and `JobError` are confusing users in relation to the `is_caller_error` attribute for `Error`. This deprecates the `is_caller_error` arguments to `ActionError` and `JobError` and replaces them with `set_is_caller_error_to` to better indicate that the argument sets `is_caller_error` on the `Error`.
- Add `stacklevel` to most deprecation warnings to provide better indication of where deprecated things are being used.

All of these changes are completely backwards compatible, but the bits that have been deprecated will be removed in PySOA 1.0.0.

### DIFF
--- a/pysoa/client/client.py
+++ b/pysoa/client/client.py
@@ -34,6 +34,12 @@ from pymetrics.instruments import TimerResolution
 from pymetrics.recorders.base import MetricsRecorder
 import six
 
+from pysoa.client.errors import (
+    CallActionError,
+    CallJobError,
+    ImproperlyConfigured,
+    InvalidExpansionKey,
+)
 from pysoa.client.expander import (
     ExpansionConverter,
     ExpansionNode,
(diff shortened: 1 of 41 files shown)


nickwilliams-eventbrite:
>I do wonder if TransientPySOATransportError is needed though. I don't see any usages yet, and it's much harder to remove classes than add them.

Agree it's harder to remove than to add. However, on Slack, specific use cases have been put forth for this intermediate base exception. Some transport errors you just don't want to ignore (indicate misconfiguration or malusage), but transient transport errors are only for indicating situations that can resolve themselves (service not currently available). The hypothetical `CircuitBreakerOpen` exception we've been discussing, for example, would inherit from `TransientPySOATransportError`. If data on the page is non-critical, you can just catch `TransientPySOATransportError` and ignore it, but you wouldn't typically want to ignore non-transient errors.

PR #227, opened 2019-11-09
[MAJOR] Refactor ServerTestCase to not inherit from unittest.TestCase

nickwilliams-eventbrite:
``unittest.TestCase`` support in PyTest is hacky and prevents a lot of features, such as parametrization of test methods and fixtures, from working. This major change refactors ``ServerTestCase`` to inherit from ``object`` instead of ``unittest.TestCase``. The resulting class has polyfills for all of the ``self.assert*`` and ``self.fail*`` methods and raises warnings if your test class uses methods that you should migrate to other methods. If you are using test plans, this is almost certainly a breaking change. If you are just using regular ``ServerTestCase``, every attempt has been made to make this change backwards-compatible, but you may find a few tests you have to fix.

The changes are as follows:

- [MINOR] New class ``BaseServerTestCase`` contains helper methods for setting up the test service, calling actions on the service, and asserting actions runs with and without errors, all of which previously resided directly in ``ServerTestCase``.
- [MINOR] New class ``UnitTestServerTestCase`` extends both ``unittest.TestCase`` and ``BaseServerTestCase`` for when you really want to use ``unittest``-style tests. Its ``setUp`` method calls the helper method for setting up the test service.
- [MINOR] New class ``PyTestServerTestCase`` extends ``BaseServerTestCase``. Its ``setup_class`` method will issue a warning and call ``setUpClass`` if your class has that method. Its ``teardown_class`` method will issue a warning and call ``tearDownClass`` if your class has that method. Its ``setup_method`` method calls the helper method for setting up the test service and will issue a warning and call ``setUp`` if your class has that method. Its ``teardown_method`` method will issue a warning and call ``tearDown`` if your class has that method. If you currently use ``addCleanup``, it will still work but will issue a warning. All the standard ``self.assert*`` and ``self.fail*`` methods are polyfilled and should work similar to the way they previously worked, but you should endeavor to swich over to simple uses of the ``assert`` keyword, which provides better diffs in PyTest failure messages.
- [MAJOR] ``ServerTestCase`` is now an alias for ``PyTestServerTestCase`` instead of a class. If you have existing test classes that extend ``ServerTestCase``, and those tests classes do not start with the word ``Test``, **PyTest will not run them anymore**! You need to rename them to start with ``Test``.
- [MAJOR] The test plan class ``ServicePlanTestCase`` now inherits from ``PyTestServerTestCase`` instead of ``UnitTestServerTestCase`` like it previously did. Test plans behave slightly differently now. Your fixture files should all work the same way they always have, but if you have any advanced uses of setup or teardown methods, they may break.
- [PATCH] Fixes a bug in ``stub_action`` preventing it from working as a decorator in PyTest-style unit test methods (it already worked properly in ``unittest``-style unit test methods).


vartec:
> [MAJOR] ServerTestCase is now an alias for PyTestServerTestCase instead of a class. If you have existing test classes that extend ServerTestCase, and those tests classes do not start with the word Test, PyTest will not run them anymore! You need to rename them to start with Test.
> [MAJOR] The test plan class ServicePlanTestCase now inherits from PyTestServerTestCase instead of UnitTestServerTestCase like it previously did. Test plans behave slightly differently now. Your fixture files should all work the same way they always have, but if you have any advanced uses of setup or teardown methods, they may break.

I'd rather that we had a [MINOR] with these inheriting from `UnitTestServerTestCase`, and only after it has been released and communicated, then have the [MAJOR] change.

nickwilliams-eventbrite:
> I'd rather that we had a [MINOR] with these inheriting from `UnitTestServerTestCase`, and only after it has been released and communicated, then have the [MAJOR] change.
> 
> Releasing breaking change w/o transitional period will make it hard for people to make their stuff forwards compatible with the new way of writing tests in gradual way.

This is going in PySOA 1.0.0, so adapting to this change will be part of undergoing that major version bump.

PR #231, opened 2019-12-02
[PATCH] Add documentation about release roadmap

nickwilliams-eventbrite:


docs/roadmap.rst
```diff
@@ -0,0 +1,76 @@
...
+* 2.0.0 will require Conformity 2.x.
+
+
+Planned Roadmap
++++++++++++++++
+
+The following table shows the current planned roadmap for release dates and support dates.
+
++--------------+-----------------+-------------------+-------------------+---------------------+
+| Release      | Supported Patch | Release Date      | Support Ends With | End of Support Date |
```
vartec: what does the "supported patch" column mean?

nickwilliams-eventbrite: It will indicate the most recent (or last, for old version series) patch version for that given minor series. For example, if 1.0.12 is released and then 1.1.0 is released and we don't release another 1.0.x patch, then that column next to 1.0.x will forever say 1.0.12.

PR #236, opened 2020-01-03
[PATCH] Skip Django database connection cleanup in unit tests

nickwilliams-eventbrite:
This should prevent the following error:

```
RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
```

### DIFF
--- a/pysoa/test/server.py
+++ b/pysoa/test/server.py
@@ -45,6 +45,7 @@ import six
 
 from pysoa.client.client import Client
 from pysoa.common.errors import Error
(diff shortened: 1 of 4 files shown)


pysoa/test/server.py
```diff
@@ -138,6 +139,11 @@ def setup_pysoa(self):  # type: () -> None
                 },
             },
         )
+        # noinspection PyProtectedMember
```
bar: why are we adding pycharm annotation to our code? 🙃

PR #238, opened 2020-01-27
[MINOR] Add ability to extend context in request.call_local_action

nickwilliams-eventbrite:


pysoa/server/types.py
```diff
@@ -161,14 +163,19 @@ def call_local_action(self, action, body, raise_action_errors=True, is_caller_er
         action_type = server.action_class_map[action]  # type: ActionType
         action_callable = action_type(server.settings)
 
+        new_context = copy.copy(self.context)
```
colink-eb: I'm curious about this copy. This means changes to the new dictionary won't change the one passed in, but changes to sub-objects will touch the objects in the original context. Is that correct/expected/accounted for?

nickwilliams-eventbrite: This doesn't permit the changing of sub-keys. Only top-level keys are affected by `new_context.update`.

Issue #255, opened 2020-06-11
TestPlans: emptydict and emptylist are ignored when global inputs are defined


leonardoa-eb:
- Your operating system name and version.

> ProductName:	Mac OS X
ProductVersion:	10.15.5
BuildVersion:	19F101

- Your Python interpreter type and version.

> Python 3.7.7

- Any details about your local setup that might be helpful in troubleshooting.
None
- Detailed steps to reproduce the bug.
Create a test plan using a global input using a nested list/dict structure, like:
```
echo.0: global input:                  global_echo_list.0.foo: bar

test name: case_001
test description: testing
# one
echo.0: expect: no errors
# two
echo.1: input emptydict:               local_echo_list.0:
echo.1: input:                         local_echo_list.1.bar: foo
echo.1: input emptydict:               local_echo_list.2:
echo.1: expect: no errors
``` 
And this test case:
```
from __future__ import absolute_import, unicode_literals
import os

from pprint import pprint as pp
from conformity import fields

from pysoa.server.action.base import Action
from pysoa.server.server import Server
from pysoa.test.plan import ServicePlanTestCase

class EchoAction(Action):
    request_schema = fields.SchemalessDictionary()
    response_schema = fields.SchemalessDictionary()
    def run(self, request):
        pp(request.body)
        return request.body

class StubServer(Server):
    service_name = 'stub'
    action_class_map = {'echo': EchoAction}

class TestPySOAPlans(ServicePlanTestCase):
    server_class = StubServer
    server_settings = {}
    fixture_path = os.path.dirname(__file__) + '/plans'
```

Run that test and verify that `local_echo_list` only has 1 item in it.

- Expected Behavior:
Run that test and receive a `local_echo_list` with 3 items in it.

PR #256, opened 2020-06-11
[PATCH] Fix handling of empty structures in `get_all_paths`

leonardoa-eb:
Fixes https://github.com/eventbrite/pysoa/issues/255

if [_global_directives are filled](https://github.com/eventbrite/pysoa/blob/236ac4147b9cd3ff96150e80bf2bcbcf0283c2b5/pysoa/test/plan/parser.py#L273) is used `get_all_paths` to recursively define where put each value detailed on the current test_case.
The iterable-like types are traveled for each item but for empty cases `{}` or `[]` this doesn't happen.
We'll check before doing the iteration block if are truth-ty values and if not return the current_path value.

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -682,7 +682,7 @@ class Server(object):
             threads.setdefault(thread_id, {'name': thread_id})['traceback'] = stack
 
         extra = {'data': {'thread_status': {
-            t['name']: [l.rstrip() for l in t['traceback']] for t in threads.values()
+            t['name']: [l.rstrip() for l in t['traceback']] for t in threads.values()  # noqa: E741
         }}}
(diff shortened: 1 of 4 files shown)


pysoa/server/server.py
```diff
@@ -682,7 +682,7 @@ def harakiri(self, signal_number, _stack_frame):  # type: (int, FrameType) -> No
             threads.setdefault(thread_id, {'name': thread_id})['traceback'] = stack
 
         extra = {'data': {'thread_status': {
-            t['name']: [l.rstrip() for l in t['traceback']] for t in threads.values()
+            t['name']: [l.rstrip() for l in t['traceback']] for t in threads.values()  # noqa
```
leonardoa-eb, line 685: I did it just because I want to satisfy the 'linter'

PR #257, opened 2020-06-12
[MINOR] Allow Server subclasses to set a custom introspection action

sethbrite:
Currently, it is more difficult than it needs to be to add information to the `introspect` action responses. This change allows a service implementer to easily override the default introspection action.

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -153,6 +153,9 @@ class Server(object):
     service_name = None  # type: Optional[six.text_type]
     action_class_map = {}  # type: Mapping[six.text_type, ActionType]
 
+    # Allow a server to specify a custom introspection action
+    introspection_action = None  # type: ActionType
+
     def __init__(self, settings, forked_process_id=None):
         # type: (ServerSettings, Optional[int]) -> None
         """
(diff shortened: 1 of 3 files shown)


sethbrite:
testme

pysoa/server/server.py
```diff
@@ -557,8 +561,12 @@ def execute_job(self, job_request):  # type: (EnrichedJobRequest) -> JobResponse
                 if action_in_class_map:
                     action = self.action_class_map[action_request.action](self.settings)
                 elif action_request.action == 'introspect':
-                    from pysoa.server.action.introspection import IntrospectionAction
-                    action = IntrospectionAction(server=self)
+                    # If set, use custom introspection action. Use default otherwise.
+                    if self.introspection_action is not None:
+                        action = self.introspection_action(server=self)
+                    else:
+                        from pysoa.server.action.introspection import IntrospectionAction
```
akirmse-eb, line 568: I've run into problems with local imports like this, because they're not run every time the code is.  e.g. I encountered a broken import that hadn't been found because it was in a branch for an error case.  Is there any harm in importing this at the top level?

sethbrite: Unfortunately, in that case `pysoa.server.server` and `pysoa.server.action.introspection` would import each other. This causes an import loop and the `pysoa.server.server` module fails to load. This could be solved by moving (a substantial amount) of code around in PySOA, but it would also break some import paths.

PR #261, opened 2020-07-17
Capture stubs to autogenerate schemas for contract testing

eb-nahuel:
Register stub responses and requests in stub_action for use in prototype contract testing

### DIFF
--- a/pysoa/server/server.py
+++ b/pysoa/server/server.py
@@ -690,7 +690,7 @@ class Server(object):
             threads.setdefault(thread_id, {'name': thread_id})['traceback'] = stack
 
         extra = {'data': {'thread_status': {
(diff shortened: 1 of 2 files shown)


pysoa/server/server.py
```diff
@@ -690,7 +690,7 @@ def harakiri(self, signal_number, _stack_frame):  # type: (int, FrameType) -> No
             threads.setdefault(thread_id, {'name': thread_id})['traceback'] = stack
 
         extra = {'data': {'thread_status': {
-            t['name']: [l.rstrip() for l in t['traceback']] for t in threads.values()  # noqa: E741
```
gabrielf-eb, line 693: I found this lint error, and I fixed it. I didn't like to omit the error by adding the exception # noqa

PR #262, opened 2020-07-22
[MINOR] Treat UTC datetime strings as datetimes in test plans

leonardoa-eb:
# Description

As part of feature parity with our deprecated library of test-plans, we need to support for date-times represented as strings for `expect any` grammar ([related code in Legacy SOA](https://github.com/eventbrite/soa/blob/7.34.4/service/test/test_plan/data_types.py#L75-L99)).

### DIFF
--- a/tests/unit/test/plan/grammar/test_data_types.py
+++ b/tests/unit/test/plan/grammar/test_data_types.py
@@ -46,6 +46,11 @@ class TestAnyValue(unittest.TestCase):
         self.assertEqual(copy.copy(data_types.AnyValue('str')), 'Hello')
         self.assertEqual(copy.deepcopy(data_types.AnyValue('str')), 'Hello')
 
+    def test_datetime(self):
+        self.assertEqual(data_types.AnyValue('datetime'), '2018-09-01T00:00:00Z')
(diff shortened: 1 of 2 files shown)


tests/unit/test/plan/grammar/test_data_types.py
```diff
@@ -46,6 +46,11 @@ def test_copy(self):
         self.assertEqual(copy.copy(data_types.AnyValue('str')), 'Hello')
         self.assertEqual(copy.deepcopy(data_types.AnyValue('str')), 'Hello')
 
+    def test_datetime(self):
+        self.assertEqual(data_types.AnyValue('datetime'), '2018-09-01T00:00:00Z')
+        self.assertNotEqual(data_types.AnyValue('datetime'), 'Hello')
+        self.assertNotEqual(data_types.AnyValue('datetime'), 15)
+
```
jeff-eb, line 53: datetime is accounted for, but shouldn't there be tests for `date` as well?

leonardoa-eb: You are right! Working on that!

leonardoa-eb: Added @jeff-eb . Thanks for your review

PR #264, opened 2020-08-04
[MINOR] Pass instance index into transport.

akirmse-eb:
This will allow the transport to know which forked process it is,
which is necessary if transport instances are sharing resources
like ports.

### DIFF
--- a/pysoa/common/transport/base.py
+++ b/pysoa/common/transport/base.py
@@ -119,7 +119,17 @@ class ClientTransport(Transport):
 class ServerTransport(Transport):
     """
     The base server transport defining the interface for transacting PySOA payloads on the server side.
+
+    :param service_name: The name of the service for which this transport will receive requests and send responses
+    :param instance_index: The 1-based index of this process among multiple forks
+    :param metrics: The optional metrics recorder
     """
+
+    def __init__(self, service_name, instance_index, metrics=noop_metrics):
+        # type: (six.text_type, int, MetricsRecorder) -> None
+        super(ServerTransport, self).__init__(service_name, metrics)
+
+        self.instance_index = instance_index
 
     @abc.abstractmethod
     def receive_request_message(self):
--- a/pysoa/common/transport/base.py
+++ b/pysoa/common/transport/base.py
@@ -125,7 +125,7 @@ class ServerTransport(Transport):
     :param metrics: The optional metrics recorder
     """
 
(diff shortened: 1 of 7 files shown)


pysoa/server/server.py
```diff
@@ -182,12 +182,15 @@ def __init__(self, settings, forked_process_id=None):
                 elif publisher.get('kwargs', {}):
                     _replace_fid(publisher['kwargs'], fid)
 
+        instance_index = forked_process_id or 1
```
sethbrite, line 185: forked_process_id starts at 1?

pysoa/common/transport/base.py
```diff
@@ -119,8 +119,18 @@ def receive_response_message(self, receive_timeout_in_seconds=None):
 class ServerTransport(Transport):
     """
     The base server transport defining the interface for transacting PySOA payloads on the server side.
+
+    :param service_name: The name of the service for which this transport will receive requests and send responses
+    :param instance_index: The 1-based index of this process among multiple forks
+    :param metrics: The optional metrics recorder
     """
 
+    def __init__(self, service_name, instance_index, metrics=noop_metrics):
```
sethbrite, line 128: This is technically a breaking change to add a new required positional argument. We'll want to make sure we note this in the changelog for others that have created their own `ServerTransport` subclasses.

pysoa/server/server.py
akirmse-eb: yep: https://github.com/eventbrite/pysoa/blob/master/pysoa/server/server.py#L164

pysoa/common/transport/base.py
```diff
@@ -119,8 +119,18 @@ def receive_response_message(self, receive_timeout_in_seconds=None):
 class ServerTransport(Transport):
     """
     The base server transport defining the interface for transacting PySOA payloads on the server side.
+
+    :param service_name: The name of the service for which this transport will receive requests and send responses
+    :param instance_index: The 1-based index of this process among multiple forks
+    :param metrics: The optional metrics recorder
     """
 
+    def __init__(self, service_name, instance_index=1, metrics=noop_metrics):
```
sethbrite, line 128: One tiny nit: can you put `instance_index` after `metrics`? Changing the order is still technically breaking for code that instantiates ServerTransport classes providing `metrics` as a positional argument. Putting `instance_index` after `metrics` will fix that potential issue. Other than that, looks good.

PR #266, opened 2021-03-31
[PATCH] Add tin number to sensitive fields

eb-fcarabelli:
TIN NUMBER should be mask as it is sensitive information

### DIFF
--- a/pysoa/common/logging.py
+++ b/pysoa/common/logging.py
@@ -188,6 +188,7 @@ class RecursivelyCensoredDictWrapper(object):
         'personal_id_numbers', 'personal-id-numbers', 'personalIdNumbers',
         'personal_identification_number', 'personal-identification-number', 'personalIdentificationNumber',
         'personal_identification_numbers', 'personal-identification-numbers', 'personalIdentificationNumbers',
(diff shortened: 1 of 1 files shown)


pysoa/common/logging.py
```diff
@@ -188,6 +188,7 @@ class RecursivelyCensoredDictWrapper(object):
         'personal_id_numbers', 'personal-id-numbers', 'personalIdNumbers',
         'personal_identification_number', 'personal-identification-number', 'personalIdentificationNumber',
         'personal_identification_numbers', 'personal-identification-numbers', 'personalIdentificationNumbers',
+        'tin_number', 'tin-number', 'tinNumber',
```
anunley-eb, line 191: `tinNumber` is akin to `ATM Machine` (Automatic Teller Machine Machine).  I'd recommend we just spell it out like we do with personal identification number?

anunley-eb (approved):
I mean, I _still_ think the field itself should be `tax_identification_number` but that ship sailed a while back.  Naming things continues to be the hardest thing in software engineering.  🥲

Diff looks fine!

PR #268, opened 2021-06-08
[PATCH] Limit the backoff time on retries

eb-enzo:
Limit the backoff time on retries so it doesn't exceeds the timeout
setting for the request.

The default of 10 retries can block the sender for up to around 128
seconds. As the default timeout is 5 seconds, it the request should
timeout quite before that.

### DIFF
--- a/pysoa/common/transport/redis_gateway/core.py
+++ b/pysoa/common/transport/redis_gateway/core.py
@@ -347,7 +347,22 @@ class RedisTransportCore(object):
             # Try at least once, up to queue_full_retries times, then error
             for i in range(-1, self.queue_full_retries):
                 if i >= 0:
-                    time.sleep((2 ** i + random.random()) / self.EXPONENTIAL_BACK_OFF_FACTOR)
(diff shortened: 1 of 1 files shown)


pysoa/common/transport/redis_gateway/core.py
```diff
@@ -347,7 +347,22 @@ def send_message(
             # Try at least once, up to queue_full_retries times, then error
             for i in range(-1, self.queue_full_retries):
                 if i >= 0:
-                    time.sleep((2 ** i + random.random()) / self.EXPONENTIAL_BACK_OFF_FACTOR)
+                    backoff_time_in_seconds = (2 ** i + random.random()) / self.EXPONENTIAL_BACK_OFF_FACTOR
```
brutledge-eb, line 350: wonder what that random is for?

brutledge-eb (approved):
LGTM.

pysoa/common/transport/redis_gateway/core.py
eb-enzo: You don't want all the clients retrying at the same time I guess. So the random part will help will that.

eb-enzo:
As the CI is broken, I will wait until it's fixed before retrying this fix.

PR #271, opened 2022-08-18
Use Ubuntu 20.04 in Dockerfile

miguelsanchez-eb:
Tests were not working anymore because they were using a deprecated Ubuntu version (16.04).

This PR updates the Dockerfile to use the newest available (22.04 was not an option because [it does not support Python 3.5 anymore](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa)).

I needed to make some adjustments to make everything work with the new Ubuntu version:

In the Dockerfile:
- `ENV DEBIAN_FRONTEND=noninteractive` to avoid showing [this prompt](https://askubuntu.com/questions/909277/avoiding-user-interaction-with-tzdata-when-installing-certbot-in-a-docker-contai)
- `python3.6-distutils` and `python3.7-distutils` seems to not be core python packages anymore in new Ubuntu versions.
- Regular `get-pip.py` doesn't support Python 3.6 anymore, you need to fetch it separately

Some other changes:
- `mypy<=0.910`. I ran it locally and it start complaining above some of the pysoa files with versions above that one.
- "mysql:5.7-debian" because the default one now is Oracle's, which does not have apt-get.
- Local build in the docker-compose file. Since docker-compose 1.25.1 docker will display a warning when trying to get a remote image, so I needed to explicit state that the build is local. _[Docker Compose now reports images that cannot be pulled, however, are required to be built.](https://docs.docker.com/compose/release-notes/#1251)_

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ test_plan_requirements = test_helper_requirements + [
 ]
 
 mypy_require = [
-    'mypy~=0.740;python_version>"3.4"',
+    'mypy~=0.740,<=0.910;python_version>"3.4"',
     'types-six~=0.1.7;python_version>"3.4"',
     'types-setuptools~=57.0.0;python_version>"3.4"',
     'types-mock~=0.1.3;python_version>"3.4"',
--- a/tests/functional/docker/Dockerfile-mysql
+++ b/tests/functional/docker/Dockerfile-mysql
@@ -1,4 +1,4 @@
-FROM mysql:5.7
+FROM mysql:5.7-debian
 
 RUN apt-get update && apt-get install -y locales locales-all tzdata && apt-get clean
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:20.04
 
-RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
-    apt-get install -y \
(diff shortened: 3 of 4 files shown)


setup.py
```diff
@@ -50,7 +50,7 @@ def readme():
 ]
 
 mypy_require = [
-    'mypy~=0.740;python_version>"3.4"',
+    'mypy~=0.740,<=0.910;python_version>"3.4"',
```
miguelsanchez-eb, line 53: Our code doesn't pass all the checks for newer versions of mypy, so pysoa CI is failing again. I think it's worth to limit the mypy version to make the CI work again, as it's not trivial to update the underlying code for every check (it deals with low level python like ctypes.pointer, etc.).

tests/functional/docker/Dockerfile-mysql
```diff
@@ -1,4 +1,4 @@
-FROM mysql:5.7
+FROM mysql:5.7-debian
```
eb-octavio, line 1: why debian?

eb-octavio: You already explain this in the summary of the PR.

eb-octavio (approved):
This change looks sane to me. All the changes seem to be required to make the build to be correct and pass.

PR #272, opened 2022-08-29
Make PySOA compatible with newer versions of attrs

miguelsanchez-eb:
Make PySOA compatible with newer versions of attrs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
       script:
         - tox -e py35,py35-pyinotify,py35-pytest51,py35-pytest53,coverage
     - stage: build
-      name: "Unit Tests: Python 3.7 with Attrs 18 and 19; PyTest 4.6 and 5.3"
+      name: "Unit Tests: Python 3.7 with Attrs 18 to 21; PyTest 4.6 and 5.3"
       python: "3.7"
       script:
         - tox -e py37-attrs18,py37-attrs19,py37-attrs20,py37-attrs21,py37-pytest46,py37-flake8,coverage,py37-mypy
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -32,20 +32,20 @@ matrix:
       script:
         - ./functional.sh verbose
     - stage: build
-      name: "Unit Tests: Python 2.7 with Attrs 18 and 19"
```
miguelsanchez-eb, line 35: I realized there was a bug here, the comment said we were testing with attrs 19 but we weren't. I've fixed it and included 2 more versions of attrs.

.travis.yml
```diff
@@ -32,20 +32,20 @@ matrix:
...
       script:
-        - tox -e py27-attrs18,py27-flake8,py27-coverage
+        - tox -e py27-attrs18,py27-attrs19,py27-attrs20,py27-attrs21,py27-flake8,py27-coverage
     - stage: build
       name: "Unit Tests: Python 3.5 with and without PyInotify; Pytest 5.1 and 5.3"
       python: "3.5"
       script:
         - tox -e py35,py35-pyinotify,py35-pytest51,py35-pytest53,coverage
     - stage: build
-      name: "Unit Tests: Python 3.7 with Attrs 18 and 19; PyTest 4.6 and 5.3"
```
miguelsanchez-eb, line 45: I realized there was a bug here, the comment said we were testing with attrs 19 but we weren't. I've fixed it and included 2 more versions of attrs.

tox.ini
```diff
@@ -19,6 +19,8 @@ deps =
     .[testing]
     attrs18: attrs~=18.2
     attrs19: attrs~=19.1
+    attrs20: attrs~=20.3
+    attrs21: attrs~=21.3
```
eb-octavio, line 23: doesn't this allow for version `21.4` which is the undesired one?

miguelsanchez-eb: You're right, I've fixed it.

PR #275, opened 2022-12-21
[MINOR] EB-207224 -- fixed unicode serialization issue

rmathews-eb:
This PR should fix [incident 617](https://eventbrite.slack.com/archives/C04FMN96BDY) that is being caused due to invalid utf-8 characters being returned from some service calls.

### DIFF
--- a/pysoa/common/serializer/msgpack_serializer.py
+++ b/pysoa/common/serializer/msgpack_serializer.py
@@ -94,7 +94,7 @@ class MsgpackSerializer(BaseSerializer):
         if not isinstance(data_dict, dict):
             raise ValueError('Input must be a dict')
         try:
-            return msgpack.packb(data_dict, default=self._default, use_bin_type=True, unicode_errors='replace')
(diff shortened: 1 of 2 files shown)


pysoa/common/serializer/msgpack_serializer.py
```diff
@@ -94,7 +94,7 @@ def dict_to_blob(self, data_dict):  # type: (Dict) -> six.binary_type
         if not isinstance(data_dict, dict):
             raise ValueError('Input must be a dict')
         try:
-            return msgpack.packb(data_dict, default=self._default, use_bin_type=True)
+            return msgpack.packb(data_dict, default=self._default, use_bin_type=True, unicode_errors='replace')
```
luca-eb, line 97: we should use `surrogateescape` to tackle the specific error we are receiving. this library is used by ALL pysoa services so using `replace` might give us undesired effects elsewhere (where for instance we want the unicode encoding to fail)

luca-eb (changes requested):

luca-eb (approved):

PR #276, opened 2023-01-04
Fixing pipeline issues.

rmathews-eb:


pysoa/common/transport/redis_gateway/backend/base.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, redis_connection):  # type: (redis.StrictRedis) -> None
 
         :param redis_connection: The connection for registering this script
         """
-        self._redis_script = redis_connection.register_script(self._script.strip())  # type: redis.client.Script
+        self._redis_script = redis_connection.register_script(self._script.strip())  # type: ignore
```
rmathews-eb, line 41: This change was due to the fact that for some reason mypy couldn't find the `redis.client.Script` type and was throwing an error on it.

mypy.ini
```diff
@@ -0,0 +1,3 @@
+[mypy]
+implicit_optional = True
+ignore_missing_imports = True
```
rmathews-eb, line 3: Configuration here squashed a majority of the mypy errors

tox.ini
```diff
@@ -45,7 +45,7 @@ commands = flake8
 
 [testenv:py37-flake8]
 skip_install = true
-deps = flake8~=3.7,>=3.7.8
+deps = flake8~=5.0
```
rmathews-eb, line 48: The version of flake being used was throwing an error when run. `5.0` appears to work for python 3.7

tests/unit/common/test_serializers.py
```diff
@@ -204,7 +204,7 @@ def test_currint(self, value):
 
     def test_invalid_utf8(self):
         serializer = MsgpackSerializer()
-        input_value =  {'v': u'value \ud83d'}
+        input_value = {'v': u'value \ud83d'}
```
rmathews-eb, line 207: Flake fix leftover from my previous PR where flake never ran in the pipeline.

liamcassidy-eb (approved):

shamlet-eb (approved):

pysoa/common/transport/redis_gateway/backend/base.py
rmathews-eb: I think what happened is the type annotation was probably added while mypy wasn't actually being run due to travis CI being down (or someone just merged with a breaking build since they don't seem to be blocking PRs). All tests still pass and as mypy shouldn't affect runtime I think we should be ok. Is there another level of testing you had in mind for us to run on this?

eb-octavio: I would say this must be tested in QA at least. Bumping the new version on your service, then build, then deploy.

rmathews-eb: Definitely will do! This is going into soa_proxy_service so I will work with Luca and others to make sure that everything is verified there.

